PR #36, opened 2017-09-14
Implement broadcasting for numbers

edubart:
This PR extend broadcasting for numbers too, the broadcasting is made using a tensor of data size 1 and strides 0, so it should be memory and performance efficient

Some tests from broadcasting where also added which was missing


mratsim:
Indeed, I forgot to implement tests for broadcasting.

Thanks for that. I will add more tests soon so that + - * / work fine as well with broadcasting.

PR #40, opened 2017-09-14
In-place division/multiplication by scalar, closes #20

### LINKED ISSUE #20: Add inplace multiplication and division by a scalar
`+=` and `-=` are done

`*=` and `/=` are missing

edubart:
This also introduces the fmap_inplace function which is used to implement those operations.


mratsim:
For the inplace map/fmap I was thinking of using `apply` as a name but I'm still confused between apply and applymap as it's commonly used in other data science package (pandas ...).

I'll sleep on it. In any case, `fmap_in_place` is so descriptive that it won't be a problem to release it and eventually deprecate it.

edubart:
Renamed to `apply`.

PR #47, opened 2017-09-16
Fix a mistake in apply

edubart:
In the other PR I've pasted the map() code without attention, so apply wasn't working correctly with reshaped tensors, this commit fixes it


mratsim:
Thanks,

Can you add a test case with a slice?

PR #48, opened 2017-09-17
Fix apply and add its test

edubart:
The added test case was failing before and is succeeding now


edubart:
The other PR https://github.com/mratsim/Arraymancer/pull/47 github automatically closed somehow when I did push the new commit

mratsim:
Thank you

PR #50, opened 2017-09-19
Optimize element accessors

edubart:
Micro optimization in tensor element accessors

Using zip here increases an additional cost of its `min()` call, there is no need to use zip iterator in this section as `check_index(idx)` already checks for `idx` to match the rank of the tensor, meaning that both `idx` and `t.strides` should have the same length.

Also inline are added.

Adding inline I got in increased ~24% performance, removing zip usage went to 27%.

Much of the improvement here was inlining, using inlines in critical sections does make a great difference in performance.

My tests were doing a convolution forward-pass all with `--passC:"-march=native" -d:blis -d:release`
in the forward pass I do excessively use accessors to rearrange the matrix.

### DIFF
--- a/src/arraymancer/accessors.nim
+++ b/src/arraymancer/accessors.nim
@@ -19,22 +19,22 @@ proc check_index(t: Tensor, idx: varargs[int]) {.noSideEffect.}=
                     ", is different from tensor rank: " &
                     $(t.rank))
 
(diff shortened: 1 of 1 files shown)


mratsim:
Excellent, thanks.
I really hope inline iterators chaining is solved ASAP (https://github.com/nim-lang/Nim/issues/4516 and [forum](https://forum.nim-lang.org/t/2972))

PR #53, opened 2017-09-19
Fix bug in inplace division by scalar, add tests

edubart:
Division by scalar should have been separated for integers and real numbers

Tests were missing also added


mratsim:
Thanks, good catch

PR #56, opened 2017-09-20
Fix as type with sliced tensors, add tests

edubart:
`astype` was failing with sliced tensors because it was using `map` directly on tensor data, fixed and added tests


mratsim:
I see the bug now, converting the whole data is fine, but strides and offset should be copied over.

Well doing a map and converting the Tensor to a contiguous one is an upfront cost but it should be paid back later anyway.

Issue #51, opened 2017-09-19
Add tests for shallowSlicing, unsafeView (both proc), shallowReshape, unsafeReshape 


mratsim:

mratsim:
Fixed: https://github.com/mratsim/Arraymancer/commit/ec0c65db82c3e9a0b14e75cf2deb54c0bc6c8ef7#diff-ab75e2e8d3564c9535dbce6a75b34e4d

PR #57, opened 2017-09-20
Fix shallow/unsafe reshape on slices

edubart:
Shallow reshape function was missing to set the offset, making it fail with slices
Also added tests for this bug and for shallow/unsafe reshape in general


mratsim:
Great, thanks, I definitely need more tests. Half of https://github.com/mratsim/Arraymancer/issues/51 is done

Issue #22, opened 2017-09-08
Create a `map` CUDA kernel for element-wise operation


mratsim:
For generic elementwise operation, a map kernel is needed:

- copy, converting to contiguous
- Fill with specific or random value
- additions, etc on 3D+ tensors
- elementwise log, sin, etc on 3D+ tensors

Torch example: https://github.com/torch/cutorch/blob/1.0/lib/THC/THCApply.cuh
Neon Cuda template: https://github.com/NervanaSystems/neon/blob/v2.1.0/neon/backends/cuda_templates.py

mratsim:
Gist with code snippet to call a custom CUDA kernel from Nim:

https://gist.github.com/mratsim/dfbd944f64181727a97dffb30b8cbd0a

PR #59, opened 2017-09-22
Cuda functors

mratsim:
- Ground work for generic element-wise kernels
- Replace cublas add/sub and in-place add/sub by element-wise kernels that support non-contiguous tensors or slices of any rank


mratsim:
Practical implementation of https://github.com/mratsim/Arraymancer/issues/22

PR #60, opened 2017-09-22
Optimize strided iterator super slowdown in some cases

edubart:
Using the library I've encountered with a massive slowdown when concatenating datasets.

Test case:
```Nim
import arraymancer
import times

var s = newSeq[Tensor[int]]()
for i in 1..256:
  s.add(zeros([1,3,128,128], int))

let start = cpuTime()
let res = concat(s, 0)
echo "elapsed ", (cpuTime() - start)*1000, "ms"
```

Before this PR:

```
elapsed 15842.251ms
```

After:

```
elapsed 107.187ms
```

Tests were made in release mode, roughly 147x factor in this test.

The mais issue was the closure, which made a copy to the tensor, as the library copy tensors by value we can never pass a tensor to a closure without copying it, workaround is to pass by reference, although `unsafeAddr` must be used..

Futhermore allocation of seq were removed from the `strided_iteration` and using stack arrays with MAXDIMS size increasing more the performance in my tests, then caching strides, shape seqs in let statements at the function head improved a little more the performance.

### DIFF
--- a/src/arraymancer/accessors.nim
+++ b/src/arraymancer/accessors.nim
@@ -52,27 +52,29 @@ template strided_iteration[T](t: Tensor[T], strider: IterKind): untyped =
   ## Iterate over a Tensor, displaying data as in C order, whatever the strides.
 
   ## Iterator init
-  var coord = newSeq[int](t.rank) # Coordinates in the n-dimentional space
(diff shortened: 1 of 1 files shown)


mratsim:
Excellent, all speed improvements to this iterator are welcome, it's the most critical piece of code for CPU tensors I think.

PR #63, opened 2017-09-24
Add unsqueeze and stack, with tests

edubart:
I do set the strides to be consistent with the rest of the lib, although I'm not sure if it was really needed because this new dimension will have length 1 so we will never loop on it, but it shouldn't hurt


edubart:
I've update the PR to also add `stack` function, as it depends on `unsqueeze`

PR #66, opened 2017-09-24
Add unsafeToTensor and unsafeToTensorReshape

edubart:
Both functions are useful for loading data from disk without doing an extra copy when passing to a tensor, tests added

Also tests for toTensorReshape were added, and its function exported


mratsim:
Thanks for that. Regarding the "normal" version of toTensorReshape I'd rather not expose it publicly, the difference is just a dot between seq.toTensor.reshape([2,2]) and seq.toTensorReshape([2,2]).

For the unsafe version it's good, unsafeToTensor.unsafeReshape is really verbose.

PR #67, opened 2017-09-24
Add fold2, to interate over two tensors

edubart:
With tests, argument syntax order chosen to be consistent with the rest of the API

### DIFF
--- a/src/arraymancer/higher_order.nim
+++ b/src/arraymancer/higher_order.nim
@@ -114,6 +114,20 @@ proc fold*[U, T](t: Tensor[U],
   for val in t.axis(axis):
     result = f(result, val)
 
+proc fold2*[T](t1: Tensor[T],
+               start_val: T,
+               f: (T, T, T)-> T,
+               t2: Tensor[T]
+               ): T {.noSideEffect.}=
+  ## Chain result = f(result, element1, element2) over all elements of two Tensors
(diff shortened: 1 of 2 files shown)


mratsim:
There are 2 big things that makes me uneasy to merge `fold2`
### I. Technical part
1. Arguments order (for function call and for the closure)
    - Since the signature is (t1, start_val, t2) I would like the `f` function to accept (val1, result, val2). But this is an awkward order for a function (result, val1, val2 is better). Inversely result.fold2(t1, t2) for function call would be awkward. I don't know think order is good.

2. Instead of T, T, T, fold2 should accept T, U, V -> U (U being the type of the result)

3. Currently if the Tensors are not of the same size, iteration stops. But some people will probably prefer for it to throw an error. This means that we might have to extend fold2 with an extra optional argument.

### II. Usage and ergonomics
Beyond those, I don't remember seeing any default widespread functions that are ternary.
This means that everyone will create its own custom function and plug it into fold2.

In that case it's probably easier for people to use an iterator in this style.

```Nim
zip(t1: Tensor[T], t2: Tensor[U]): tuple[a: T, b: U] =
asset t1.size == t2.size # or not

for i in zip(t1.values, t2.values):
  yield (t1, t2)
```
The benefits is that there will be no doubt of the order between result, val1, val2

PR #73, opened 2017-09-26
Optimize many critical functions

edubart:
I've profiled a lot my demos, also read the nim generated C code to do these optimizations.
The rationale is that most of tensor operations are done in contiguous tensors of the same size, when that is true we can do simple optimized for loops and skip the stride operator altogether. Nevertheless strided iteration was optimized a little too. Everything done after many tests. Before this commit my DNN benchmark performance for 200 iterations of forward and backward pass was `5089ms`, with this patch it's now `3045ms` while pytorch with similar DNN stands at 1539ms.

### DIFF
--- a/src/arraymancer/accessors.nim
+++ b/src/arraymancer/accessors.nim
@@ -56,34 +56,42 @@ type
 template strided_iteration[T](t: Tensor[T], strider: IterKind): untyped =
   ## Iterate over a Tensor, displaying data as in C order, whatever the strides.
 
-  ## Iterator init
-  var iter_pos = t.offset
-  let rank = t.rank
-  let shape = t.shape
-  let strides = t.strides
-  var coord: array[MAXRANK, int]
-  var backstrides: array[MAXRANK, int]
-  for i in 0..<rank:
-    backstrides[i] = strides[i]*(shape[i]-1)
+  block:
+    let size = t.shape.product
 
-  ## Iterator loop
-  for i in 0 .. <shape.product:
-
-    ## Templating the return value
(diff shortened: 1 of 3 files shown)


mratsim:
The changes to data_structure and accessors are good.

Instead of changing higher_order I would prefer if we directly change the iterators like this (Note it fails a contiguous conversion test):

```Nim
template element_iteration[T](t: Tensor[T], strider: IterKind, need_contiguous: bool): untyped =
  let contig_check =  if not need_contiguous: true
                                  elif t.isContiguous and need_contiguous: true
                                  else: false

  if t.data.len == t.size and contig_check:
    for i in 0..<t.data.len:
      yield t.data[i]
  else:
    strided_iteration(t, strider)

iterator items*[T](t: Tensor[T]): T {.noSideEffect.}=
  ## Inline iterator on Tensor values
  ##
  ## The iterator will iterate in C order regardingless of the tensor properties (Fortran layout, non-contiguous, slice ...).
  ## So [0, 0, 0] then [0, 0, 1] then ... then [0, 1, 0] ...
  ##
  ## Usage:
  ##  .. code:: nim
  ##     for val in t: # items is implicitly called
  ##       val += 42
  t.element_iteration(IterKind.Values, true)
```
**If we can't it's fine.**

Also regarding map2 and apply2:
  1. Even if contiguous they should have the same C, F order, otherwise you have to iterate one in order and the other in a transposed order (which requires strided iteration)
  2. If they have the exact same shapes and strides we can also don't worry about the strided iteration: caveat if the Tensor is big it's a tradeoff between useless computation and cache prediction/prefetching.

PR #74, opened 2017-09-26
Use openmp for higher order functions

edubart:
Introduced `omp_parallel_countup` template which does use OpenMP to parallelize for loops above a threshold set in global config.

By default openmp is disabled, to enable it user have to compile with --passC:"-fopenmp" -passL"-fopenmp", to control the number of threads the user can do `export OMP_NUM_THREADS=4` otherwise OpenMP will auto detect the number of threads.

Note: I did a replace all in the source code removing all {.noSideEffect.}, too many functions was 
dependent on high order functions, and also because of laziness.. Does noSideEffect really improve performance for nim?


Yardanico:
@edubart noSideEffect has nothing to do with performance, see https://nim-lang.org/docs/manual.html#pragmas-nosideeffect-pragma

edubart:
Appveyor script needs to be updated I think, it's failing

edubart:
Already merged.

PR #79, opened 2017-09-27
OpenMP for strided operations

edubart:
I will leave this here to mark progress

The idea is to do strided operations via parallel blocks. Each block iterates over a different range.

- [x] map
- [x] apply
- [x] map2
- [x] apply2
- [ ] fold?
- [ ] fold2?


edubart:
I do not plan to paralelize map or reduce soon, this PR can be merged once reviewed.

PR #81, opened 2017-09-27
[WIP] Conv2D and nnpack bindings

edubart:
Here I'm implementing 2d forward and backward low level functions, with nnpack and fallback implementation.

The PR includes NNPack bindings generated with c2nim with various hand fixes and an additional interface for using it.

To enable NNPack must compile with `-d:nnpack`, user may want to install nnpack with scalar backend if they do not have a high end Intel CPU.

The nn module is completely independent from arraymancer tensor core (see that I haven't changed any arraymancer source, instead only added files), this can be a good practice because not everyone may use all the arraymancer interfaces.

To use user has to `import arraymancer/nn`

### DIFF
--- a/src/nn_primitives/backend/nnpack.nim
+++ b/src/nn_primitives/backend/nnpack.nim
@@ -1,0 +1,486 @@
+# Copyright 2017 the Arraymancer contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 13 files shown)


edubart:
I've added low level `conv2d_forward` and `conv2d_backward` both works with NNPack and the fallback implementation. The fallback implementation is based on CS231n class and uses `im2col` + matrix multiply for forward and backward. To check the validity of the algorithm, forward/backward on random tensor is done in the tests and compared the numerical gradient using the relative error.

edubart:
I've refactored the code, added more documentation and more tests.

edubart:
Rebased with latest changes.

Issue #71, opened 2017-09-25
Explore write-tracking/escape analysis to avoid copying seq to closure


mratsim:
Not implemented: https://nim-lang.org/docs/manual.html#effect-system-read-write-tracking

Reference ticket: https://github.com/nim-lang/Nim/issues/3377

> Solution 3.
> In the example above we could mark s as lifted variable. This would change the signature of giveMeProc from giveMeProc(s: seq[int]): proc() to giveMeProcs: mustBeOnTheHeap seq[int]): proc(). We also need to keep track of the variables: the giveMeProc(@[1,2,3]) is a call with parameter which is in static, therefore we need to copy it before the function call.
> This solution is not that easy to implement and also has some bad implications. In the following case:
> 
> proc wrapper(s: seq[int]): proc() =
>   result = giveMeProc(s)
> we need to realize that the parameter of wrapper needs to be on the heap. Moreover, if at the implementation at wrapper we might not now the actual implementation of giveMeProc, therefore we actually need to introduce an additional keyword to the language (to indicate that the parameter s needs to be on the heap at forward declarations).

> Solution 3 is essentially:
> 
> proc wrapper(s: seq[int]): proc () {.escapes: [s].}
> 
> That's why I told you to look at my writetracking. ;-)
> Escapes: [s] means you cannot pass a static seq to it.

mratsim:
Solved by https://github.com/mratsim/Arraymancer/pull/90

Issue #23, opened 2017-09-08
Use inline iterators to avoid copying


mratsim:
Unfortunately inline iterator chaining is broken: https://github.com/nim-lang/Nim/issues/4516
As a workaround, closures are currently used

Occurrences:
- https://github.com/mratsim/Arraymancer/blob/8551acb3561878c6effc04b5f39ee211e5f36031/src/arraymancer/operators_blas_l1.nim#L58
- https://github.com/mratsim/Arraymancer/blob/087b5c82addf90fd3a569364613e89b683f011a1/src/arraymancer/accessors_slicer.nim#L461
- https://github.com/mratsim/Arraymancer/blob/8551acb3561878c6effc04b5f39ee211e5f36031/src/arraymancer/operators_blas_l1.nim#L50
- https://github.com/mratsim/Arraymancer/blob/ab2bca5d772c054e71ac4ac1aaaa493655be149d/src/arraymancer/comparison.nim#L19
- https://github.com/mratsim/Arraymancer/blob/68a9c2bb4e7e84680446d5435ffb21f9896dcead/src/arraymancer/operators_extra.nim#L26

mratsim:
Alternative, use template:

```Nim
zip(a, b).mapIt(it[0] + it[1])
```

mratsim:
Alternative resolution by https://github.com/mratsim/Arraymancer/pull/90

PR #90, opened 2017-10-06
[WIP] Rework iterators

edubart:
This is a completely rework for iterators and closures. The ideia is to get rid of all closures, use templates for map and apply, and improve performance.

Improvements:
- Improves the performance by a good margin (~10% in my dnn demo)
- Solves #23 by not needing inline iterators
- Depracates #71 by not using closures
- Implements enumerate mentioned in the TODOs of the source as `pairs`
- Implements `mpairs` mentioned the TODOs of the source
- Implements `mzip` mentioned the TODOs of the source
- Make the compiler able to vectorize loops
- Make paralelization with openmp easier
- Reimplement all Blas L1 operators without closures
- Removed unwatend allocations from iteration (that closures generated)
- Avoid one extra `is_C_Contiguous` call in `map` and `apply`
- Introduce dual strided iterator, my solution for not using closures or inline iterators, the dual strided iterator is optimized to handle 4 cases, if t1,t2 is contiguous, if either t1 or t2, or if none are
- Fixed rare crash with openmp (it did crash sometimes because of closures allocation were allocating and we cant allocate while inside openmp threads)

Side effects:
- `reduce` removed for a while, we have fold already, do we need it? can implement later
- some docs were removed in the rework process because I was trying to code fast, also the api changed, will add later
- `pairs` works different now it returns the index not the coords, which is much useful and is the mentioned enumerate, also the old pairs did allocate a new seq every iteration which was very slow, also I don't see the need to iterate with coords, however can implement again later, implementing it again should be done without seqs
- 2 test related with the above was disable for now
- lots of iterators removed and new ones introduced
- `real_indices` removed as it not needed anymore
- `values` removed as `items` already does the same

New iterators:
- `items` `mitems` (like before)
- `pairs` `mpairs` (its the enumerate)
- `zip`  `mzip` (new zip without closures)
- `zipPairs` `mzipPairs` (zip with enumerate)

New high order, element-wise without closures, they work like the other ones, but use templates:
- `eapply`
- `eapply2`
- `eapply3`
- `emap`
- `emap2`

All tests looks green


mratsim:
Awesome work. A few remarks about the side effects:

Side effects:
- `reduce` removed for a while, we have fold already, do we need it? can implement later
> yes it's needed, reducing is a very common operation. Removing it would mean people would need to think "I want to multiply all element, I should use 1 as a starting point. If I need max, I should use low(float32) as a starting point".
> In any case I plan to use OpenMP parallel reduction for common operations [supported by OpenMP](https://computing.llnl.gov/tutorials/openMP/#REDUCTION) (min, max, +, * ...). like so `#pragma omp parallel for shared(sum, a) reduction(+: sum)`, a being `tensor.data`

- some docs were removed in the rework process because I was trying to code fast, also the api changed, will add later
> No problem

- `pairs` works different now it returns the index not the coords, which is much useful and is the mentioned enumerate, also the old pairs did allocate a new seq every iteration which was very slow, also I don't see the need to iterate with coords, however can implement again later, implementing it again should be done without seqs
> The previous `pairs` was meant as an external API so that people can just get the coordinates and not interact at a low-level with `tensor.data` and strides.

> I prefer that we keep it like that and introduce a "items_and_memOffset" or "items_with_real_index" iterator. It's fine if `pairs` is slow or use the previous `strided_iteration` implementation. I don't have a use case for it that isn't covered by map/apply/apply2, etc

- 2 test related with the above was disable for now
> No problem
- lots of iterators removed and new ones introduced
> Public ones that were in v2.0 should be deprecated
- `real_indices` removed as it not needed anymore
> No problem
- `values` removed as `items` already does the same
> For now it should at most be deprecated, but it is probably interesting to still expose a closure iterator and just stating that they are slower than the inline iterator. Reduce can be kept with `values` for the moment.

Is there a reason to export `emap`, `eapply`, etc? I'd rather have only the common map/apply.
Also they should be called `mapT`, `applyT` to keep in line with how internal template are called.

Anyway, thank you, that's awesome !

edubart:
> Is there a reason to export emap, eapply, etc? I'd rather have only the common map/apply.
Also they should be called mapT, applyT to keep in line with how internal template are called

`emap` and `eapply` should be encouraged to be used instead of normal `map` and `apply` why? because by not having a closure gcc can generate vectorized codes compatible with gcc auto vectorization see https://gcc.gnu.org/projects/tree-ssa/vectorization.html which is the intention of this PR

althought full vectorizaton with SIMD does still need using `__builtin_assume_aligned` gcc can at least infer that the pointers are aligned to the `sizeof(ptrdiff_t)` and make some vectorized assembly calls 

`emap` `eapply` generates code in a simple loop without closures/functions call, while `map` e `apply` per design makes call a function pointer making auto vectorization impossible (and it's impossible change this behavior, inlining a function in map/apply does not make sense because we need to pass its pointer)

edubart:
To demonstrates the above, go to https://gcc.godbolt.org/
place the following code:

```C
//#define ALIGNMENT 32
//#define NO_CLOSURE

static int __attribute__ ((noinline)) closure_add(int a, int b, int& c)
{
	c = a + b;
}

int square(int *ia, int* ib, int *ic, int num) {
	int *a, *b, *c;
	#ifdef ALIGNMENT
		a = (int*)__builtin_assume_aligned(ia, ALIGNMENT);
		b = (int*)__builtin_assume_aligned(ib, ALIGNMENT);
		c = (int*)__builtin_assume_aligned(ib, ALIGNMENT);
	#else
		a = ia;
		b = ib;
		c = ic;
	#endif
    for(int i=0;i<num;++i)
	#ifdef NO_CLOSURE
		c[i] = a[i] + b[i];
	#else
		closure_add(a[i], b[i], c[i]);
	#endif
}
```

Use gcc 7.2 with flags `-O3 -march=native`, `-O3` enables auto vectorization.

Play with it, uncomment `define NO_CLOSURE` define I see the vectorization ocurr this is what happens when using `emap` instead of `map`, the generated code is similiar with this example (checked C generated sources of nim).

Not uncomment `define ALIGNMENT 32` see vectorization algorhtming improving with less generated asm code. *TIP: you can hover mouse to `vmovdqa` for example to see what this instruction is doing.*

edubart:
Recent changes:
* `pairs` works like before (with coords), the new behavior was moved to `enumerate`
* added removed docs back and some comments
* reduce reimplemented
* enable the 2 disabled tests
* make `reshape` `asContiguous` parallel by using emap/eapply
* use the new `emap` and `eapply` in u_`func` and `math_functions`
* remove a lot of inline, 

Why remove inlines? emap/eapply does generate a huge function in C because it contains many cases for the strided iteratations becoming a huge function, inlining simple operators like + - would mean unfolding a huge function with lots of branches and loops every time, so we should not inline anything with emap/eapply to reduce binary code size, and the gains with inline for them it's insignificant, inline does help with small functions like `isContiguous`

This concludes my intention here, I will leave any function naming change or additional documentation up to you.

What's important now is that no API compability was broken (I think).

Futher enhacements (for other PRs):
* remove closure iterator in the axis iterator
* make reduce/fold parallel as mentioned
* make operations from slices read/mutations use emap/eapply instead of for loops with zip, we should replace all for loops operations with emap/eapply because they can work in parallel, at least in the core manually fors should not be used because we always want to explore parallelism

PR #96, opened 2017-10-08
Use import instead of include

mratsim:
## Pros:
- Compilation times are shorter because touching one file does not trigger the whole compilation
- Nimsuggest doesn't work with include because it cannot follow dependencies not specified in the current file. (This may save battery due to nimsuggest crash/locks)

## Cons:
- GCC does not see the whole program. Note: Link-time optimization `-flto` improves arraymancer-demos by 20% anyway (at a very long compilation cost) so irrelevant for max perf. Also Nim dead-code elimination/template/inline works fine with imports.
- Potentially long import lines in each file like `../../tensor/tensor` in a nested hierarchy for nn layers for example.


mratsim:
Appveyor-Github integration was broken but build passing: https://ci.appveyor.com/project/mratsim/arraymancer/build/338

PR #98, opened 2017-10-08
Fix rare crashes while using openmp

edubart:


src/arraymancer/accessors.nim
```diff
@@ -119,10 +119,9 @@ template initStridedIteration(coord, backstrides, iter_pos: untyped, t, iter_off
   if iter_offset != 0:
     var z = 1
     for i in countdown(t.rank - 1,0):
-      let z2 = z*t.shape[i]
-      coord[i] = (iter_offset div z) mod z2
+      coord[i] = (iter_offset div z) mod t.shape[i]
       iter_pos += coord[i]*t.strides[i]
-      z = z2
+      z *= t.shape[i]
```
edubart: The previous code here was not working for all cases, it's an important fix on the iterator, this code is only used when openmp is enabled

src/arraymancer/shapeshifting.nim
```diff
@@ -68,7 +68,8 @@ template contiguousT[T](result, t: Tensor[T], layout: OrderType): untyped=
   if layout == rowMajor:
     result = t.mapT(x)
   else:
-    result = t.transpose().mapT(x)
```
edubart: t.transpose() was being called inside `omp_parallel_blocks` template which caused crashes, also transpose did non needed extra copy

PR #103, opened 2017-10-10
Rework axis iterator and parallelize reduce 

edubart:
This PR objectives are:

- [x] Rework axis iterator without closures
- [x] Remove extra copies from axis iterator
- [x] Add versions of reduce with fused ops
- [x] Make reduce parallel with OpenMP
- [x] Optimize sum/mean using the above

### DIFF
--- a/arraymancer.nimble
+++ b/arraymancer.nimble
@@ -105,6 +105,7 @@ task test_native, "Run all tests - march=native":
 
 task test_openmp, "Run all tests - OpenMP":
   switch("define","openmp")
+  switch("stackTrace","off")
   test "all_tests"
 
 task test_mkl, "Run all tests - Intel MKL - single threaded":
--- a/src/tensor/accessors.nim
+++ b/src/tensor/accessors.nim
@@ -16,6 +16,32 @@ import  ./private/p_accessors,
         ./private/p_checks,
         ./data_structure
 
+proc atContiguousIndex*[T](t: Tensor[T], idx: int): T {.noSideEffect,inline.} =
+  ## Return value of tensor at contiguous index
+  ## i.e. as treat the tensor as flattened
+  return t.data[t.getContiguousIndex(idx)]
+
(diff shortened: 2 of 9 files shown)


edubart:
Fold was not touched because there is no much sense in folding with multiple start values in parallel, also user may use fold expecting to fold sequentially.

New public functions introduced:
- `atContiguousIndex` returns value at C contiguous index
- `atAxisIndex` and `unsafeAtAxisIndex` returns tensor with dimension collapsed to 1 in the specified axis and axis's index
- `omp_parallel_reduce_blocks` introduced to make paralell reduce implementation easy, this was somewhat tricky to implement, there are a lot of injects and templates going on
- `reduceT` and `reduceAxisT` reduce with fused operations and openmp-aware 

The two new public functions `atContiguousIndex` and `atAxisIndex` was needed to implement reduce in different offsets, they are also public to the make templates work.

Using a different name to reduce on axis was needed to avoid errors while compiling, I was not able to make both named just `reduceT`

Changes:
- Many missing bounds checks were added to iterators to make easy to catch bugs while implementing
- Axis iterator completely reworked
- Sum reworked
- Some noSideEffect removed to make compile ok in OpenMP
- The old reduces does use `reduceT`

There is a new global config `OMP_MAX_REDUCE_BLOCKS` this is needed to use static arrays while reducing, avoiding allocation of seqs, currently set to 8, meaning that at max will use 8 threads to reduce and allocate 8 variables to reduces on the function stack.

I got huge gains in all my benchmarks that were using `sum`.

For example of usage see `aggregate.nim`

```Nim
proc sum*[T: SomeNumber](t: Tensor[T]): T =
  ## Compute the sum of all elements of T

  t.reduceT():
    x+=y

proc sum*[T: SomeNumber](t: Tensor[T], axis: int): Tensor[T] =
  ## Compute the sum of all elements of T along an axis
  t.reduceAxisT(axis):
    x+=y
```

Here `x` is mutable to avoid extra copies, so it works different from `applyT` and `mapT`,  meaning in `reduceT` we should mutate instead of returning a value.

PR #115, opened 2017-10-13
Min, max, variance and standard deviation

edubart:
Implement some aggregate statistics functions

- [x] min
- [x] min over axis
- [x] max
- [x] max over axis
- [x] variance
- [x] variance over axis
- [x] standard deviation
- [x] standard deviation over axis

All of them were implemented with parallelism in mind. To make them in parallel, `fold_inline` and `fold_inline_axis` was implemented, these functions work something similar to fold but to make them able work in parallel the function takes 3 bodies, one for initializing, one to folding in parallel, and last one to merge folds, so is quite flexible. Check variance implementation to see how it works.

Tests implemented for all of them. Also checked with openmp.

### DIFF
--- a/src/tensor/aggregate.nim
+++ b/src/tensor/aggregate.nim
@@ -110,12 +110,12 @@ proc variance*[T: SomeReal](t: Tensor[T], axis: int): Tensor[T] {.noInit.} =
     x += y
   result /= (t.shape[axis]-1).T
 
-proc standard_deviation*[T: SomeReal](t: Tensor[T]): T {.inline.} =
+proc std*[T: SomeReal](t: Tensor[T]): T {.inline.} =
   ## Compute the standard deviation of all elements
   ## The normalization is by the (n-1), like in the formal definition
   sqrt(t.variance())
 
-proc standard_deviation*[T: SomeReal](t: Tensor[T], axis: int): Tensor[T] {.noInit,inline.} =
+proc std*[T: SomeReal](t: Tensor[T], axis: int): Tensor[T] {.noInit,inline.} =
   ## Compute the standard deviation of all elements
   ## The normalization is by the (n-1), like in the formal definition
   sqrt(t.variance(axis))
--- a/src/tensor/aggregate.nim
+++ b/src/tensor/aggregate.nim
@@ -80,8 +80,9 @@ proc max*[T](t: Tensor[T], axis: int): Tensor[T] {.noInit.} =
       ex = max(ex,ey)
 
 proc variance*[T: SomeReal](t: Tensor[T]): T =
-  ## Compute the variance of all elements
-  ## The normalization is by the (n-1), like in the formal definition
+  ## Compute the sample variance of all elements
+  ## The normalization is by (n-1), also known as Bessel's correction,
+  ## which partially correct the bias of estimating a population variance from a sample of this population.
   let mean = t.mean()
   result = t.fold_inline() do:
     # Initialize to the first element
--- a/src/tensor/aggregate.nim
+++ b/src/tensor/aggregate.nim
@@ -31,6 +31,16 @@ proc sum*[T](t: Tensor[T], axis: int): Tensor[T] {.noInit,noInit.} =
   t.reduce_axis_inline(axis):
     x+=y
 
+proc product*[T](t: Tensor[T]): T {.inline.}=
+  ## Compute the product of all elements
+  t.reduce_inline():
+    x*=y
+
+proc product*[T](t: Tensor[T], axis: int): Tensor[T] {.noInit,inline.}=
+  ## Compute the product along an axis
+  t.reduce_axis_inline(axis):
+    x.melwise_mul(y)
+
 proc mean*[T: SomeInteger](t: Tensor[T]): T {.inline.}=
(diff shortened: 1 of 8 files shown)


mratsim:
Nice !

Have to read on `do:` magic

I put that on hold until the C++ fix is ready has I will have to rebase and create an alternate C++ path for mzip

edubart:
Added:

- [x] product
- [x] product over an axis
- [x] clamp (inplace too)
- [x] element-wise mul without broadcasting (inplace too)
- [x] element-wise div without broadcasting (inplace too)

edubart:
Added generating random tensor in the normal distribution.

src/tensor/aggregate.nim
```diff
@@ -14,26 +14,98 @@
...
+
+proc min*[T](t: Tensor[T]): T =
+  ## Compute the min of all elements
+  t.reduce_inline():
+    x = min(x,y)
+
+proc min*[T](t: Tensor[T], axis: int): Tensor[T] {.noInit.} =
+  ## Compute the min along an axis
+  t.reduce_axis_inline(axis):
+    for ex, ey in mzip(x,y):
```
mratsim: Can you add the alternate cpp code path

src/tensor/aggregate.nim
```diff
@@ -14,26 +14,98 @@
...
+
+proc max*[T](t: Tensor[T]): T =
+  ## Compute the max of all elements
+  t.reduce_inline():
+    x = max(x,y)
+
+proc max*[T](t: Tensor[T], axis: int): Tensor[T] {.noInit.} =
+  ## Compute the max along an axis
+  t.reduce_axis_inline(axis):
+    for ex, ey in mzip(x,y):
```
mratsim: here too

src/tensor/aggregate.nim
```diff
@@ -14,26 +14,98 @@
...
+proc variance*[T: SomeReal](t: Tensor[T], axis: int): Tensor[T] {.noInit.} =
+  ## Compute the variance of all elements
+  ## The normalization is by the (n-1), like in the formal definition
+  let mean = t.mean(axis)
+  result = t.fold_axis_inline(axis) do:
+    # Initialize to the first element
+    x = square(y - mean)
+  do:
+    # Fold in parallel by summing remaning elements
+    for ex, ey, em in mzip(x,y,mean):
```
mratsim: here too

src/tensor/aggregate.nim
```diff
@@ -14,26 +14,98 @@
...
+  do:
+    # Fold in parallel by summing remaning elements
+    x += square(y - mean)
+  do:
+    # Merge parallel folds
+    x += y
+  result /= (t.size-1).T
+
+proc variance*[T: SomeReal](t: Tensor[T], axis: int): Tensor[T] {.noInit.} =
+  ## Compute the variance of all elements
```
mratsim: Can you change to "Compute the sample variance of all elements"

src/tensor/aggregate.nim
```diff
@@ -14,26 +14,98 @@
...
+  ## The normalization is by the (n-1), like in the formal definition
```
mratsim: I would say "The normalization is by (n-1), also known as Bessel's correction, which partially correct the bias of estimating a population variance from a sample of this population"

src/tensor/aggregate.nim
```diff
@@ -14,26 +14,98 @@
...
+    for ex, ey, em in mzip(x,y,mean):
+      ex += square(ey - em)
+  do:
+    # Merge parallel folds
+    x += y
+  result /= (t.shape[axis]-1).T
+
+proc standard_deviation*[T: SomeReal](t: Tensor[T]): T {.inline.} =
+  ## Compute the standard deviation of all elements
+  ## The normalization is by the (n-1), like in the formal definition
```
mratsim: same thing as for the variance

src/tensor/higher_order.nim
```diff
@@ -66,25 +66,46 @@ template map3_inline*[T](t1, t2, t3: Tensor[T], op:untyped): untyped =
...
   omp_parallel_reduce_blocks(reduced, block_offset, block_size, t.size, 1, op) do:
-    t.atContiguousIndex(block_offset)
+    x = t.atContiguousIndex(block_offset)
   do:
     for y {.inject.} in t.items(block_offset, block_size):
       op
   reduced
 
-template reduceAxisT*[T](t: Tensor[T], axis: int, op: untyped): untyped =
+template fold_inline*[T](t: Tensor[T], op_initial, op_middle, op_final: untyped): untyped =
```
mratsim: 👍

src/tensor/aggregate.nim
mratsim: It won't be needed. I found a way to isolate C and C++.

src/tensor/aggregate.nim
mratsim: It won't be needed. I found a way to isolate C and C++.

src/tensor/aggregate.nim
mratsim: It won't be needed. I found a way to isolate C and C++.

PR #125, opened 2017-10-23
[WIP] CuDNN convolution on 4D tensors

mratsim:


mratsim:
Slicing is broken. Potential sources: Cuda 9, nimcuda not being updated for devel or BackwardsIndex:




`unsafeReshape` also errors out. It might be the `=` assignation that doesn't work properly

Issue #134, opened 2017-10-29
macOS OpenMP critically slow in specific circumstances


mratsim:
[Benchmark file.](https://github.com/mratsim/Arraymancer/blob/853e4f01b14f8c9a3a3851c8c998e098506fd6d5/benchmarks/implementation/logsumexp.nim)

Results on i5-5257U (dual-core mobile 2.7GHz, turbo 3.1)
Batch size 256

Compilation flag no OpenMP
`nim c -d:native -d:release --out:bin/logsumexp --nimcache:./nimcache benchmarks/implementation/logsumexp.nim`

Compilation flag OpenMP
`nim c d:openmp -d:native -d:release --out:bin/logsumexp --nimcache:./nimcache benchmarks/implementation/logsumexp.nim`

| Implementation | Classes | OpenMP | epochTime (seconds)      |
| -------------- | ------- | ------ | ------------------------ |
| Classic        | 1000    | No     | 1.869992971420288        |
| Streaming      | 1000    | No     | 1.463253974914551        |
| Classic        | 3       | No     | 0.005753040313720703     |
| Streaming      | 3       | No     | 0.004420995712280273     |
| Classic        | 1000    | Yes    | **2.416302919387817**    |
| Streaming      | 1000    | No*    | 1.478353023529053        |
| Classic        | 3       | Yes    | **0.005849123001098633** |
| Streaming      | 3       | No*    | 0.004467964172363281     |

No*: the streaming logsumexp is compiled with but does not use OpenMP

mratsim:
Confirmed on linux as well including on the `integer_matmul` benchmark.
Could it be GCC 7 or a nim.cfg issue?

mratsim:
git bisect returns this commit https://github.com/mratsim/Arraymancer/commit/218aa7bb0b412903847b8761cf6c266c53fc300d:

on MacOS

    Move OpenMP parallel section to inner loop in GEMM fallback: fix https://github.com/mratsim/Arraymancer/issues/123

Can't reproduce this on linux.

mratsim:
Further research shows that this is not isolated:

OpenMP forum:
- [OpenMP slower on OSX](http://forum.openmp.org/forum/viewtopic.php?f=3&t=1583)
- [Fast on Linux Slow on OSX](http://forum.openmp.org/forum/viewtopic.php?f=3&t=1017)

Stack overflow:

[...]

PR #139, opened 2017-10-30
OpenMP: Fix false sharing in parallel reduce and fallback GEMM (especially on macOS)

mratsim:
Unfortunately PR https://github.com/mratsim/Arraymancer/pull/137 had data races in GEMM.

After reverting, using a chunksize bigger than a 64kb cache line (16 for float32, 8 for int64, etc) i.e. setting `schedule(static, 16)` didn't help.
This solution is suggested to combat false sharing, when a cache line is modified it's invalidated for all the other thread, generating huge slowdown not unlike what was detected.

Resources:
- [Use of OpenMP chunk to break cache](https://stackoverflow.com/questions/18342033/use-of-openmp-chunk-to-break-cache)
- [False sharing in OpenMP loop array access](https://stackoverflow.com/questions/45032586/false-sharing-in-openmp-loop-array-access)
- Paper: [Detect false sharing in OpenMP application](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.221.7676&rep=rep1&type=pdf)

The false sharing issue might remain because the microkernel is 16 elements anyway so cache line sized.

Alternative would be to parallelize at a higher level. According to BLIS here are the suggestion:
 - BLIS retreat: [Parallelizing GEMM within BLIS](https://www.cs.utexas.edu/users/flame/BLISRetreat/BLISRetreatTalks/BLIS_RETREAT_TYLER.pdf)
 - Paper:[ Opportunities for parallelism in matrix multiplication](https://www.cs.utexas.edu/users/flame/pubs/FLAWN71.pdf)
- Blis [wiki on multithreading](https://github.com/flame/blis/wiki/Multithreading)

Loop around microkernel
image

5th loop | BLIS_JC_NT | n |  
4th loop | N/A | k | Not enabled
3rd loop | BLIS_IC_NT | m |  
2nd loop | BLIS_JR_NT | n |  
1st loop | BLIS_IR_NT | m

>When compute resources have private L3 caches (example: multi-socket systems), try parallelizing the JC loop. This means threads (or thread groups) will pack and compute with different row panels from matrix B.

> For compute resources that have private L2 caches but that share an L3 cache (example: cores on a socket), try parallelizing the IC loop. In this situation, threads will share the same packed row panel from matrix B, but pack and compute with different blocks of matrix A.

> If compute resources share an L2 cache but have private L1 caches (example: pairs of cores), try parallelizing the JR loop. Here, threads share the same packed block of matrix A but read different packed micro-panels of B into their private L1 caches. In some situations, parallelizing the IR loop may also be effective.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ The library is inspired by Numpy and PyTorch.
   - [Full documentation](#full-documentation)
   - [Features](#features)
     - [Speed](#speed)
+      - [Parallelism](#parallelism)
+      - [Memory allocation](#memory-allocation)
       - [Micro benchmark: Int64 matrix multiplication](#micro-benchmark-int64-matrix-multiplication)
       - [Logistic regression](#logistic-regression)
       - [DNN - 3 hidden layers](#dnn---3-hidden-layers)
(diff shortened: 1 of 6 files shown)


mratsim:
Another performance bottleneck was found in the parallel reduce code due to false sharing.

See http://www.drdobbs.com/parallel/eliminate-false-sharing/217500206

The results array which stores the intermediate results from each thread will now take into account the number of items per cache-line so that threads to do not compete for writing the same cache-line.

Unfortunately we can't use `sizeof` at compile-time to determine if only 8 (int64/float64) or 16 (int32/float32) or 0 (Tensor) elements can fit in a cache-line so we use an arbitrary 16 values.

See feature request: https://github.com/nim-lang/Nim/issues/5802
And PR: https://github.com/nim-lang/Nim/pull/5664

This fixes: https://github.com/mratsim/Arraymancer/issues/134

PR #165, opened 2017-12-05
Fix nimble command in README.md

dwd31415:
Change `nimble arraymancer` to the correct version `nimble install arraymancer`


mratsim:
Thanks !

PR #173, opened 2017-12-17
[Docs] fix overlap of ')' into 2nd col of Tensor / CUDATensor tab

Vindaar:
While looking at your r/nim post about 0.3.0 (nice!), I noticed a small issue in the docs. In the table comparing the features of Tensors and CudaTensors, the closing parenthesis of one row overlaps into the 2nd column.

Just thought I'd fix that real quick.


mratsim:
Good catch, thanks

PR #196, opened 2018-03-17
Improved Euler benchmarks

narimiran:
Several changes to all three versions to make them faster.

EDIT: now I see you were also working on those files. Let me see if I can resolve the conflicts.
EDIT2: should be ok now.


mratsim:
Awesome thank you

Issue #197, opened 2018-03-17
Superfluous `_` in slicing is hard to find (broken ex02)


bluenote10:
I tried running `ex02_handwritten_digits_recognition.nim` but I got an error during accuracy testing:

```
Epoch #0 done. Testing accuracy
ex02_handwritten_digits_recognition.nim(124) ex02_handwritten_digits_recognition
p_accessors_macros_read.nim(36) slicer
p_checks.nim(67)         check_steps
Error: unhandled exception: Your slice start: 0, and stop: -1, or your step: 1, are not correct. If your step is positive
                start must be inferior to stop and inversely if your step is negative
                start must be superior to stop. [IndexError]
```

I think I have an idea what is going wrong, but it took quite some time to find the issue: I guess the problem is simply that the shape of `y_test` is one dimensional, i.e., the `_` in the expression `y_test[i ..< i+1000, _]` is probably wrong. I will provide a small PR to fix that...

Would it make sense to add another sanity check which would produce a clear error message when there is such a superfluous index?

mratsim:
Yes you can add a new check, thanks!

PR #198, opened 2018-03-17
fixed slicing in ex02

bluenote10:
Partial fix for #197

I think the slicing in the accuracy check was also not quite as intended: Before, the loop `i in 0 ..< 10` would compute the accuracy on `0..<1000`, `1..<1001`, `2..<1002`, i.e., the test samples have a huge overlap. I changed the slicing to use disjoint blocks.

I'm wondering if the expression `i*1000 ..< (i+1)*1000` could be reused, because it has to be repeated 4 times. Is there a way to construct a stand-alone `SteppedSlice`? Simply writing `let test_slice = i*1000 ..< (i+1)*1000` will pick `..<` from system, so that doesn't work. Would be nice to have a special stand-alone constructor for Arraymancer slices like:

```nimrod
let test_slice = slice(i*1000 ..< (i+1)*1000) # macro similar to []
```

### DIFF
--- a/examples/ex02_handwritten_digits_recognition.nim
+++ b/examples/ex02_handwritten_digits_recognition.nim
@@ -120,10 +120,10 @@ for epoch in 0 ..< 5:
     var score = 0.0
     var loss = 0.0
     for i in 0 ..< 10:
-      let y_pred = X_test[i ..< i+1000, _].model.value.softmax.argmax(axis = 1).indices.squeeze
-      score += accuracy_score(y_test[i ..< i+1000, _], y_pred)
+      let y_pred = X_test[i*1000 ..< (i+1)*1000, _].model.value.softmax.argmax(axis = 1).indices.squeeze
+      score += accuracy_score(y_test[i*1000 ..< (i+1)*1000], y_pred)
 
-      loss += X_test[i ..< i+1000, _].model.sparse_softmax_cross_entropy(y_test[i ..< i+1000, _]).value.data[0]
+      loss += X_test[i*1000 ..< (i+1)*1000, _].model.sparse_softmax_cross_entropy(y_test[i*1000 ..< (i+1)*1000]).value.data[0]
     score /= 10
     loss /= 10
     echo "Accuracy: " & $(score * 100) & "%"


mratsim:
Thanks! It doesn't break in release mode and the non-release was too slow so i didn't see that :/

Actually you can use that: `let slice = SteppedSlice(a: i*1000, b: (i+1) * 1000 - 1, step: 1)`

Or `let slice = i*1000..<(i+1)*1000` the macro accepts normal slices.

bluenote10:
> Or let slice = i*1000..<(i+1)*1000 the macro accepts normal slices.

I tried that, but that doesn't seem to work:

```
Error: type mismatch: got <Tensor[system.int], HSlice>
but expected one of: 
proc slicer[T](t: AnyTensor[T]; slices: varargs[SteppedSlice]): AnyTensor[T]
  first type mismatch at position: 2
  required type: varargs[SteppedSlice]
  but expression 'slice' is of type: HSlice
proc slicer[T](t: AnyTensor[T]; slices: varargs[SteppedSlice]; ellipsis: Ellipsis): AnyTensor[
    T]
  first type mismatch at position: 2
  required type: varargs[SteppedSlice]
  but expression 'slice' is of type: HSlice
proc slicer[T](t: AnyTensor[T]; ellipsis: Ellipsis; slices: varargs[SteppedSlice]): AnyTensor[
    T]
  first type mismatch at position: 2
  required type: Ellipsis
  but expression 'slice' is of type: HSlice
proc slicer[T](t: AnyTensor[T]; slices1: varargs[SteppedSlice]; ellipsis: Ellipsis;
              slices2: varargs[SteppedSlice]): AnyTensor[T]
  first type mismatch at position: 2
  required type: varargs[SteppedSlice]
  but expression 'slice' is of type: HSlice
proc slicer[T](t: Tensor[T]; slices: ArrayOfSlices): Tensor[T]
  first type mismatch at position: 2
  required type: ArrayOfSlices
  but expression 'slice' is of type: HSlice

expression: slicer(y_test, slice)
```

But I discovered that `let slice = i*1000..<(i+1)*1000 | 1` works because it forces the conversion of the second argument to `Step` resulting in calling the proper `..<` variant.

I was thinking about introducing a `slice` macro mainly because of the `desugar` stuff that is happening in the `[]` macro as well.

PR #199, opened 2018-03-17
added rank check for slicer; should fix #197

### LINKED ISSUE #197: Superfluous `_` in slicing is hard to find (broken ex02)

bluenote10:
This produces an error message when the number of expressions in a slicer exceeds the tensor rank. For example: `Error: unhandled exception: Rank of slice expression (2) is larger than the tensor rank (1). [IndexError]`

Since the examples sometimes use less expressions in the slicer than the tensor rank, I assume that is allowed?

### DIFF
--- a/src/tensor/init_cpu.nim
+++ b/src/tensor/init_cpu.nim
@@ -180,7 +180,7 @@ proc randomTensor*[T:SomeReal](shape: varargs[int], max: T): Tensor[T] {.noInit.
   ##      - a tensor backend
   ## Result:
   ##      - A tensor of the input shape filled with random value between 0 and max input value
-  randomTensorCpu(result, shape, max)
+  randomTensorCpu(result, shape, max-1)
 
 proc randomTensor*(shape: varargs[int], max: int): Tensor[int] {.noInit.} =
   ## Creates a new int Tensor filled with values between 0 and max-1.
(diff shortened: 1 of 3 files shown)


bluenote10:
`nimble test` was working fine on my branch locally. Merging in the latest master breaks the end-to-end test, which is strange, because the changes seem unrelated.

bluenote10:
`tests/end_to_end/examples.nim` should work now in release mode, but note that it also fails in non-release mode:

```
[Suite] End-to-End: Examples compile and run
examples.nim(40)         examples
linear.nim(100)          linear
ag_data_structure.nim(97) forward
linear.nim(30)           forward
nnp_linear.nim(37)       linear
operators_blas_l2l3.nim(95) *
operators_blas_l2l3.nim(63) gemm
p_checks.nim(117)        check_matmat

    Unhandled exception: Number of columns in the first matrix: 1, must be the same as the number of rows in the second matrix: 2
  [FAILED] Example 1: XOR Perceptron
```

mratsim:
Thanks,

Yes that's the side effect of `random` being deprecated and `rand` being inclusive, actually the end-to-end is still wrong :P

PR #200, opened 2018-03-18
More tweaking of Euler benchmarks

narimiran:
* changed constants to more meaningful values
* `tensor_optim` was giving wrong output - quickfix (back to the old way of assigning the boundary condition)
* less assigning in `nativearray` version, making it even faster

### DIFF
--- a/benchmarks/euler_nativearray.nim
+++ b/benchmarks/euler_nativearray.nim
@@ -34,8 +34,9 @@ proc main() =
 
   for t in 0 ..< timeSteps:
     Ts[t][0] = startingTemp - oscillations * sin(2.0 * PI * t.float / timeSteps)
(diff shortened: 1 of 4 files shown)


mratsim:
You can get even more speed by using `newSeqUninitialized` introduced in https://github.com/nim-lang/Nim/issues/6401 ;)

Otherwise it takes 1~2 second to zero the 4GB memory, and you will overwrite it again after.

narimiran:
> You can get even more speed by using `newSeqUninitialized`

Unfortunately: _Only available for numbers types._

---

Btw, I haven't spent much time debugging `tensor_optim` version, so I replaced your `apply_inline` with a for loop. If you find a way to use `apply_inline` or something else that might increase speed (while producing correct results), let me know.

PR #201, opened 2018-03-18
fix random tensor

bluenote10:
I fixed the issue yesterday by making the `randomTensor` keep its old behavior (sticking to the docstring). I have now changed it back to the breaking behavior and adjusted the docstring. There are more occurrences of `randomTensor` throughout the codebase, but in some cases it is not clear to me which convention is correct (probably it does not matter in some cases).


mratsim:
Thanks, yes it should be inclusive like everything in stdlib.
Otherwise you couldn't get high(int64) with the previous `random`.

PR #205, opened 2018-03-23
Tensor csv export

bluenote10:
I have created a `to_csv` tensor exporter for debugging purposes. It can handle tensors of arbitrary rank. Maybe this is useful in general.


.gitignore
```diff
@@ -1,3 +1,7 @@
+*
+!**/
+!*.*
+
```
bluenote10: I hope you don't mind adding these "standard" Nim gitignores which allow to keep in-place binaries (files without a dot) without seeing untracked files.

bluenote10:
In addition I have written a small Python visualization script which allows to plot the tensor contents (from the exported CSVs) into a series like this:

epoch_0_fully_connected_1
epoch_0_fully_connected_2
epoch_1_fully_connected_1
epoch_1_fully_connected_2
epoch_2_fully_connected_1
epoch_2_fully_connected_2

Currently it only supports 2D tensors, but 1D would be easy to add and the value histograms would work for other tensors as well. Are you interested in adding this e.g. under a folder `tools`?

.gitignore
mratsim: usually I just put everything in my build/ folder

mratsim:
Awesome, great addition!

For the visualization tool it might be best in it's own repo for now.
I was also thinking of continuing the [Visdom](https://github.com/facebookresearch/visdom) wrapper started by @edubart [here](https://github.com/edubart/arraymancer-vision) to add Plot.ly like capabilities to Arraymancer.

bluenote10:
It's maybe a bit too tiny for an individual repo. For now I can just share it here:

```python
#!/usr/bin/env python

from __future__ import print_function

import matplotlib.pyplot as plt
from mpl_toolkits.axes_grid1 import make_axes_locatable

import pandas as pd
import argparse


def parse_args():
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "file",
        nargs="+",
        help="CSV files to plot",
    )
    parser.add_argument(
        "-i",
        dest="interactive",
        help="Shows plot interactively",
        action="store_true",
        default=False,
    )
    return parser.parse_args()


def plot(f, args):
    try:
        df = pd.read_csv(f)
        print("Plotting file: {}".format(f))
    except pd.parser.CParserError:
        print("Failed to load file: {}".format(f))
        return

    tensor_rank = len(df.columns) - 1

    fig, axes = plt.subplots(nrows=1, ncols=2, figsize=(12, 6))

    if tensor_rank == 2:
        df_pivot = df.pivot(index="dimension_1", columns="dimension_2", values="value")

        # value histogram
        raw_values = df_pivot.values.flatten()
        axes[0].hist(raw_values, bins=70)

        # heatmap
        im = axes[1].imshow(df_pivot, aspect="auto", interpolation="none")
        divider = make_axes_locatable(axes[1])
        cax = divider.append_axes('right', size='5%', pad=0.05)
        fig.colorbar(im, cax=cax)

        fig.suptitle(f, fontsize=20)
        if args.interactive:
            plt.show()
        fig.savefig(f + ".png")

    else:
        print("Tensor of rank {} are not supported yet.".format(tensor_rank))


if __name__ == "__main__":
    args = parse_args()

    for f in args.file:
        plot(f, args)
```

mratsim:
oh, feel free to add it to a new tools folder then :)

PR #222, opened 2018-05-03
Tanh activation function

manguluka:


mratsim:
Almost missed that today, awesome thanks!

mratsim:
@manguluka, I didn't catch it as I was on mobile yesterday night. You missed multiplying by the gradient that was flowing back:

https://github.com/mratsim/Arraymancer/commit/987c7ea049ecfd190c3be15b58660c8b9b5e77df

I've also changed from x^2 to x*x as power is a slow function (though GCC/Clang `pow 2` should catch that)

manguluka:
@mratsim Thanks for catching that, and for the optimization

PR #236, opened 2018-05-17
Added higher dimensional toSeq exports

bluenote10:
For exporting purposes it would be nice if there was an inverse operation to `toTensor` for nested sequences.

This PR is the simplest solution to that, offering functions like `tensor.toSeq2D` etc. At first I though about a generic solution, but `tensor.toSeqGeneric(seq[seq[float]])` is inconvenient to type, and most use cases may involve tensors of a limited (low) rank anyway.


mratsim:
Poor GC :/. Thanks!

PR #237, opened 2018-05-19
[WIP] GRU layer

bluenote10:
I was curious to see if it is possible to provide a layer for the recently added nn-primitive GRU implementation (thanks for adding RNN's!). This is a first attempt. Do you think this is going in the right direction?

I'm still unsure if this is the way to go, because there might be an issue when unrolling the GRU over time: The unrolled GRU gates are supposed to share the `W3`, `U3`, `bW3`, `bU3` tensors, but as soon as the first (=last in forward) GRU step is back-propagated, it will update the weights immediately, which in turn affects the following back-propagations. I'm not an expert on RNN's, but my understanding was that the weights should only be updated when the entire unrolled network is back-propagated.


mratsim:
Yes it seems like a great start.

For the gradient accumulation, it should be:
```Nim
 if self.input.requires_grad:
    result[0] += self.dinput
  if self.hidden.requires_grad:
    result[1] += self.dhidden
  if self.W3.requires_grad:
    result[2] += self.dW3
  if self.U3.requires_grad:
    result[3] += self.dU3
  if self.bW3.requires_grad:
    result[4] += self.dbW3
  if self.bU3.requires_grad:
    result[5] += self.dbU3
```
instead of replacing I think.

##### GRU

Note that it should be called a `GRUCellGate` and the proc should be `gru_cell`.

I started on an actual GRU which accepts `[time/nb_sequences, batch, features]`, `gru_cell` is for a single timestep.

The gru would be the following at a high level
```
proc gru_forward(t: Tensor, ...): Tensor =
  result = zeros_like(t)
  for i in 0 ..< t.shape[0]:
    t.gru_cell_forward(...)
```

Obviously there are some tricky optimizations that can be done.

Unfortunately, I started a data science competition on Kaggle 2 days ago which is very time consuming so I will go back on fused GRUs when I burn out of the competition.

bluenote10:
> For the gradient accumulation, it should be ... instead of replacing I think.

That is what I would like to do, but does that work? As a return value, `result` is constructed from scratch so I would think there is no old content of these gradients. Maybe I misunderstand how `SmallDiffs` works.

bluenote10:
> Unfortunately, I started a data science competition on Kaggle 2 days ago which is very time consuming so I will go back on fused GRUs when I burn out of the competition.

Forgot to say: No problem at all, good luck with the competition :).

mratsim:
I've been rediving into GRU and started a PR on Stacked GRU layer with sequence/time support https://github.com/mratsim/Arraymancer/pull/283.

So after rediving, your PR is almost good with the following tricky parts:

1. Gradient accumulation is not needed, the autograd will take care of that: https://github.com/mratsim/Arraymancer/blob/92fb21f1310ea6a8182c326a8beee11aa3cd068f/src/autograd/ag_data_structure.nim#L178-L180

2. The layer and proc should be GRUCell and gru_cell

3. Note compared to stacked gru that gru_cell returns output + hidden state (both being equal) while stacked_gru returns output + end hidden state of all layers. This means that a stacked GRU returns a tuple and registers 2 nodes in the computational graph. I have no idea if this works yet :P.

mratsim:
I've taken the liberty to close this PR given the multiple changes I had to make over the weekend to make GRU work:

- Mostly MAX_NB_GRADS has been removed in #303, backward supports a new `Payload` type to allow multiples gradients flowing back for chunk, split and RNNs (output and hidden state gradients)

Also there are the following easy changes:

- Forward methods have been removed in #300 (they are now unexported static proc instead of dynamic methods)

- the `gru.nim` file should be renamed `gru_cell.nim`

I've added the fully fleshed GRU (almost, it's missing bidirectional support) as an example here: https://github.com/mratsim/Arraymancer/blob/master/examples/ex05_sequence_classification_GRU.nim

GRU_cell can be replicated by setting number of layers and seq_len to 1.

In any case, thank you for it. It made me think that I need to optimize the inference path (#301)

bluenote10:
Cool, thanks a lot for finishing this! I hope I'll have time to experiment with it soon.

PR #246, opened 2018-06-12
Add a transform boolean parameter to pca; update unit test.

metasyn:
Add a simple boolean parameter to the pca proc, so that users can get the principal components in addition to transforming the input data.

I updated the unit tests as well.

@mratsim Originally you suggested `static[bool]` - which I can still try to add if you want? I was not sure how to call the method appropriately with that signature, in the test.

### DIFF
--- a/src/ml/dimensionality_reduction/pca.nim
+++ b/src/ml/dimensionality_reduction/pca.nim
@@ -6,7 +6,7 @@ import
   ../../tensor/tensor, ../../linear_algebra/linear_algebra
 
 
-proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2, transform: bool = true): tuple[results: Tensor[T], components: Tensor[T]] =
+proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): tuple[results: Tensor[T], components: Tensor[T]] {.noInit.} =
   ## Principal Component Analysis (PCA)
   ## Inputs:
   ##   - A matrix of shape [Nb of observations, Nb of features]
@@ -26,13 +26,10 @@ proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2, transform: bool = true):
   let (_, eigvecs) = cov_matrix.symeig(true, ^nb_components .. ^1) # Note: eigvals/vecs are returned in ascending order
 
   # [Nb_obs, Nb_feats] * [Nb_feats, Nb_components], don't forget to reorder component in descending order
-  var
-    descending_eigvecs = eigvecs[_, ^1..0|-1]
-    pca_results = mean_centered * descending_eigvecs
+  result.components = eigvecs[_, ^1..0|-1]
+  result.results= mean_centered * result.components
 
-  return (pca_results, descending_eigvecs)
-
-proc pca*[T: SomeReal](x: Tensor[T], principal_axes: Tensor[T]): Tensor[T] =
+proc pca*[T: SomeReal](x: Tensor[T], principal_axes: Tensor[T]): Tensor[T] {.noInit.} =
   ## Principal Component Analysis (PCA) projection
   ## Inputs:
   ##    - A matrix of shape [Nb of observations, Nb of components]
--- a/src/ml/dimensionality_reduction/pca.nim
+++ b/src/ml/dimensionality_reduction/pca.nim
@@ -6,11 +6,14 @@ import
   ../../tensor/tensor, ../../linear_algebra/linear_algebra
 
 
-proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): Tensor[T] =
+proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2, transform: bool = true): Tensor[T] =
   ## Principal Component Analysis (PCA)
   ## Inputs:
   ##   - A matrix of shape [Nb of observations, Nb of features]
   ##   - The number of components to keep (default 2D for 2D projection)
+  ##   - transform: whether or not to output the tensor or components:
+  ##     If transform: output the dimensionality reduced tensor
+  ##     Else output the first n principal components feature plane
   ##
   ## Returns:
   ##   - A matrix of shape [Nb of observations, Nb of components]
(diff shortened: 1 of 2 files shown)


bluenote10:
In my opinion an API that changes the semantics of the return value depending on a switch is always a bit confusing. Why not return both and let the user decide what to use/discard?

src/ml/dimensionality_reduction/pca.nim
```diff
@@ -26,3 +29,19 @@ proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): Tensor[T] =
 
   # [Nb_obs, Nb_feats] * [Nb_feats, Nb_components], don't forget to reorder component in descending order
   result = mean_centered * eigvecs[_, ^1..0|-1]
+
+  if transform:
+    return result
+  else:
+    return eigvecs
```
mratsim: We should return `eigvecs[_, ^1..0|-1]` with the principal axes in descending order (those which explain the most variance first)

src/ml/dimensionality_reduction/pca.nim
```diff
@@ -26,3 +29,19 @@ proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): Tensor[T] =
...
+
+proc pca*[T: SomeReal](x: Tensor[T], principal_axes: Tensor[T]): Tensor[T] =
+  ## Principal Component Analysis (PCA) projection
+  ## Inputs:
+  ##    - A matrix of shape [Nb of observations, Nb of features]
+  ##    - A matrix of shape [Nb of observations, Nb of principal axes] to project on
```
mratsim: Document the order expected (descending)

src/ml/dimensionality_reduction/pca.nim
```diff
@@ -26,3 +29,19 @@ proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): Tensor[T] =
...
+  ##
+  ## Returns:
+  ##    - A matrix of shape [Nb of observations, Nb of components]
+  let mean_centered = x .- x.mean(axis=0)
+  result = mean_centered * principal_axes[_, ^1..0|-1]
```
mratsim: As we switch to descending order, `principal_axes` instead of `principal_axes[_, ^1..0|-1]`

metasyn:
Thanks for the feedback! I think I addressed everything.

src/ml/dimensionality_reduction/pca.nim
```diff
@@ -25,4 +26,19 @@ proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): Tensor[T] =
   let (_, eigvecs) = cov_matrix.symeig(true, ^nb_components .. ^1) # Note: eigvals/vecs are returned in ascending order
 
   # [Nb_obs, Nb_feats] * [Nb_feats, Nb_components], don't forget to reorder component in descending order
-  result = mean_centered * eigvecs[_, ^1..0|-1]
+  var
+    descending_eigvecs = eigvecs[_, ^1..0|-1]
+    pca_results = mean_centered * descending_eigvecs
+
+  return (pca_results, descending_eigvecs)
```
mratsim: Regarding lines: https://github.com/metasyn/Arraymancer/blob/98a2b4e4812bb3b0f2e203005871db255f216ca1/src/ml/dimensionality_reduction/pca.nim#L29-L33

Thanks for the change.

For performance this should be written:

```Nim
result.components = eigvecs[_, ^1..0|-1]
result.results = mean_centered * result.components
```

In your case you allocate once for descending_eigvecs, once for pca_result
And
  - it might be that you allocated once for (result_tuple), if GCC didn't put them adjacent in memory.
     it's only Tensor metadata but still.
  - fortunately there won't be 2 allocations for both tensors data because tensors have ref semantics.

src/ml/dimensionality_reduction/pca.nim
```diff
@@ -6,15 +6,16 @@ import
   ../../tensor/tensor, ../../linear_algebra/linear_algebra
 
 
-proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): Tensor[T] =
+proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2, transform: bool = true): tuple[results: Tensor[T], components: Tensor[T]] =
```
mratsim: Seems like I didn't add it as well, I would be grateful if you added the {.noInit.} pragma.

This has huge performance implication in Arraymancer (and for any library that returns a garbage collected type), see https://github.com/nim-lang/Nim/issues/7743. Otherwise Nim adds a useless GenericReset proc which is a complex proc to reset an ref object to prime state. Except that there is no need to reset a result value that will be reassigned/constructed in the proc.

```Nim
proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2, transform: bool = true): tuple[results: Tensor[T], components: Tensor[T]] {.noInit.}=
```

src/ml/dimensionality_reduction/pca.nim
```diff
@@ -25,4 +26,19 @@ proc pca*[T: SomeReal](x: Tensor[T], nb_components = 2): Tensor[T] =
...
 
   # [Nb_obs, Nb_feats] * [Nb_feats, Nb_components], don't forget to reorder component in descending order
-  result = mean_centered * eigvecs[_, ^1..0|-1]
+  var
+    descending_eigvecs = eigvecs[_, ^1..0|-1]
+    pca_results = mean_centered * descending_eigvecs
+
+  return (pca_results, descending_eigvecs)
+
+proc pca*[T: SomeReal](x: Tensor[T], principal_axes: Tensor[T]): Tensor[T] =
```
mratsim: {.noInit.} as well would be great.
```Nim
proc pca*[T: SomeReal](x: Tensor[T], principal_axes: Tensor[T]): Tensor[T] {.noInit.}=
```

Thanks

src/ml/dimensionality_reduction/pca.nim
metasyn: Got it. This also reminded me I forgot to remove transform from the signature, which I am also fixing

metasyn:
Alright, performance changes are in :)

mratsim:
Awesome, thanks.

PR #249, opened 2018-06-18
[WIP] Kmeans fixes

mratsim:
@metasyn I've been playing a bit with kmeans++ implementation to optimize it.

I've noticed that the `random` parameter is not forwarded: https://github.com/mratsim/Arraymancer/blob/afc0c213316b4d98002fbb118551a4de9b67e217/src/ml/clustering/kmeans.nim#L213-L247

Forwarding it to `let output = x.assign_labels(n_clusters, tol, max_iters, random)` fails the test, is that supposed to happen?

Another question: The labels are currently floats but, they should be integers, did I miss something?

Optimization TODO (for myself):
  - use `zeros` for intentionally zero'd Tensor, and `newTensorUninit` otherwise
  - use `CountTable` for counting occurences
  - use `result` value instead of an intermediate value.


mratsim:
What is the proper test for random centroid? Is that even testable ?

mratsim:
closed by #252

PR #252, opened 2018-06-23
Fix test when random is passed to kmeans

metasyn:
* update closest dist appropriately
* change return type from T to int for labels
* attempt to use result more often than an intermediate variable


mratsim:
Thanks for finishing it!

PR #256, opened 2018-07-07
replace all `SomeReal` by `SomeFloat`

Vindaar:
Replaces all occurrences of `SomeReal` by `SomeFloat`, because for some reason `SomeReal` now throws an `undeclared identifier` error.


mratsim:
SomeFloat doesn't exist on the latest stable unfortunately. There are 2 solutions:

- Fix Nim
- Add a `ArraymancerFloat` alias maps to SomeReal on 0.18 and SomeFloat on devel, or just == float32 or float64.

I think Nim should be fixed here.

mratsim:
upstreamed: https://github.com/nim-lang/Nim/pull/8239

Vindaar:
If that PR isn't merged, I'll add an alias for 0.18 to this PR.

mratsim:
Merged upstream.

PR #257, opened 2018-07-22
Add hdf5 support

Vindaar:
This PR will add support to read and write tensors to HDF5 files. It's a work in progress.

Features (to be) added for a start:
- [x] write tensor with generic name
- [x] write tensor with specific name
- [x] write tensor to specific group
- [x] read tensor
- [x] read tensor, convert to different datatype 
- [x] read tensor of given name
- [x] read tensor from location other than root group
- [x] write several tensors without closing H5 file in between calls
- [x] make use of `is_C_contiguous`
- [x] make sure tensors are contiguous before writing
- [x] write proper tests of writing -> reading back tensors 
  can probably always add more tests...

Once these are implemented potentially (in another PR):
- [ ] serialize NN model to H5
- [ ] read serialized NN model from H5
- [ ] read Keras NN models from H5 files

Regarding Keras H5 files, relevant Keras code: https://github.com/keras-team/keras/blob/master/keras/engine/saving.py

### DIFF
--- a/src/io/io_hdf5.nim
+++ b/src/io/io_hdf5.nim
@@ -1,0 +1,92 @@
+# Copyright (c) 2018 Mamy André-Ratsimbazafy and the Arraymancer contributors
+# Distributed under the Apache v2 License (license terms are at http://www.apache.org/licenses/LICENSE-2.0).
+# This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  ../tensor/tensor,
+  strutils, strformat, ospaths,
+  nimhdf5
+
+# constant which we use to keep track of how many tensors are stored
+# in a given H5 file without having to visit the whole H5 file.
+# We create an attribute on the root group with this name and increment
+# for each tensor.
+const NumTensorStored = "numTensorsStored"
+
+proc read_hdf5*[T: SomeNumber](hdf5Path: string): Tensor[T] {.noInit.} =
+  ## Reads a .h5 file (written by arraymancer) and returns a tensor of the
+  ## specified type.
+  ## If the tensor is stored in a different type in the file, it will be
+  ## converted.
+  ##
+  ## Input:
+  ##   - The path to a HDF5 file as a string
+  ## Output:
+  ##   - A tensor
+  discard
+
+proc write_hdf5*[T: SomeNumber](t: Tensor[T],
+                                hdf5Path: string,
+                                name = "",
+                                group = "") =
+  ## Exports a tensor to a hdf5 file
+  ## To keep this a simple convenience proc, the tensor is stored
+  ## in the root group of the hdf5 file under a generic name.
+  ## If the `name` argument is given, the tensor is stored under this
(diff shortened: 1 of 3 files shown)


mratsim:
Btw @Vindaar do you still consider this WIP or is this ready for review/merge?

Vindaar:
Oh, sorry. I got a little side tracked with my "Keras investigation" and then the last few weeks happened.

Reading Keras files will certainly not be part of this PR. Besides some technical issues (they're using sometimes using static and sometimes variable length strings for attributes for whatever reasons), which I fixed, I have some more general problems implementing this. 
My main issue comes down to: how should I use the Keras config, which I read at run time from a H5 file to create a NN in Arraymancer, where I'd normally use the DSL at compile time for that? Several hacky ideas later, I'm still not sure. :)

Are you happy with the `canImport` solution or should `nimhdf5` be a proper dependency? And I'll take a look at the long compilation times again later today and see if that still happens.

Vindaar:
Finally checked the compilation times. While `read_hdf5` calls are still on the slower side, compiling the hdf5 test takes about `3.3 s` on my laptop. Seems fine for me.

Two questions: 
- I just noticed I still have four `assert` statements in the `read_hdf5`. Shall I remove them? 
- Do you want me to rebase onto the current master branch?

mratsim:
The assert are fine, I need to do a pass to remove all the doAssert I did anyway and implement proper checks.

Yep rebasing would be perfect!

src/io/io_hdf5.nim
```diff
@@ -0,0 +1,191 @@
...
+  let tCont = t.asContiguous
+  # write tensor data
+  dset.unsafeWrite(tCont.get_data_ptr, tCont.size)
+
+  # now write attributes of tensor
+  dset.attrs["rank"] = tCont.rank
+  dset.attrs["shape"] = @(tCont.shape)
+  dset.attrs["size"] = tCont.size
+  # workaround since we can't write bool attributes
+  dset.attrs["is_C_contiguous"] = if tCont.is_C_contiguous: "true" else: "false"
```
mratsim: In which case is this needed? It's forced contiguous on line 159?

Vindaar: Oh, good point. I'm not entirely sure anymore what I had in mind. I guess to store whether the original tensor was contiguous or not.

Vindaar:
Oh, now I remember what I had in mind. It wasn't about whether the tensor is contiguous or not, but whether it is row or column based. 
So the idea was to store that information with the tensor so when it's read back, it's done in the same major ordering. 
My thinking was: convert tensor to contiguous, check whether it is now C or F contiguous. Write as attribute. Data is stored, read back and put into the same ordering.

See line 90 here: https://github.com/mratsim/Arraymancer/pull/257/files#diff-e2d8651a1d48176c63843ddfd120c20fR90.

mratsim:
Oh I see, I'll readd then

PR #266, opened 2018-08-14
Bugfix for bug in kmeans initialization

metasyn:
Rather silly bug I noticed today when giving a short presentation on nim.

The random value chosen should be be between 0 and n-1, rather than n (where n is the number of rows), since using n as the slice will result in an index out of bounds.

I also refactored the tests a little to make it easier to update in the future, saving the data and expected results into npy files.


mratsim:
Thank you!

Issue #185, opened 2017-12-25
[Autograd] Variables cannot be dereferenced


mratsim:
Due to the following template, variables cannot be dereferenced.

https://github.com/mratsim/Arraymancer/blob/fc4ad528f6afcd377c16ff99f19151a0f5e46f89/src/autograd/ag_accessors.nim#L18-L28

This forces to use casting instead of `addr v[]`

https://github.com/mratsim/Arraymancer/blob/fc4ad528f6afcd377c16ff99f19151a0f5e46f89/src/autograd/ag_data_structure.nim#L116-L120

Solution is probably easy. Dispatch to system.`[]` when varargs = 0.

mratsim:
Still relevant but it's OK to cast as a GC-ed pointer is compatible with a non-GC-ed one and it will likely won't change.

Issue #241, opened 2018-05-24
regression: SIGSEGV in example if wrapped in a proc


Vindaar:
Just updated Arraymancer only to realize that my code, which previously compiled just fine, suddenly throws a SIGSEGV. Did some testing to find that, e.g. [example 2](https://github.com/mratsim/Arraymancer/blob/master/examples/ex02_handwritten_digits_recognition.nim) does not compile on the current latest version of Arraymancer (https://github.com/mratsim/Arraymancer/commit/a0c1d52781c9bae9a10c1fc8ceb307374feb9a73)
That is with the current Nim devel version (https://github.com/nim-lang/Nim/commit/85b7d8fcc4032a0d95af3f25faf1cc850f7d59f5).

By wrapping the [example 2](https://github.com/mratsim/Arraymancer/blob/master/examples/ex02_handwritten_digits_recognition.nim) in a main proc, just like so: 
https://gist.github.com/Vindaar/d730b597f9dc17eeecfda5c053780acd
 I get (verbosity 2):
```
ex02_handwritten_digits_recognition.nim(80, 8) template/generic instantiation from here
ex02_handwritten_digits_recognition.nim(89, 32) template/generic instantiation from here
~/src/nim/arraymancer/Arraymancer/src/ml/metrics/accuracy_score.nim(26, 31) template/generic instantiation from here
~/src/nim/arraymancer/Arraymancer/src/tensor/ufunc.nim(21, 20) Warning: not GC-safe: '(proc (x): = result = U(x), nil)' [GcUnsafe]
    result = t.map(x => x.U)
                     ^
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

Did a git bisect through Arraymancer and the offending commit seems to be: https://github.com/mratsim/Arraymancer/commit/e609e998d663710281dbe161249a0139befa818c

If I remove code from the example above, I only get it to compile after everything between lines 60-96 is gone (https://gist.github.com/Vindaar/d730b597f9dc17eeecfda5c053780acd#file-ex02_wrapped-nim-L60-L96), i.e. line 60 is the first bad call.

mratsim:
Thanks for the report.

It seems like my first guess, that somehow we where dereferencing unintialized Variabe was wrong as even with a cast (without dereferencing Variable and Context) it still crashes:

```Nim
macro `[]`*[TT](v: Variable[TT], args: varargs[untyped]): Variable[TT] or VariableObj[TT]=
[...]

PR #270, opened 2018-08-30
[Pending upstream] Allow dereference of Variables

mratsim:
Fix https://github.com/mratsim/Arraymancer/issues/185

Pending https://github.com/nim-lang/Nim/pull/8715, upstream fix of https://github.com/nim-lang/Nim/issues/8706

This should avoid the regression https://github.com/mratsim/Arraymancer/issues/241


Clonkk:
@mratsim Is this still relevant 4 years later ?

mratsim:
Still relevant but it's OK to cast as a GC-ed pointer is compatible with a non-GC-ed one and it will likely won't change.

PR #273, opened 2018-09-04
Download and decompress mnist

### LINKED ISSUE #168: Auto-download, uncompress and return MNIST dataset
MNIST was added in https://github.com/mratsim/Arraymancer/commit/379aef3dc9cac0deac6d9f6f2c603643ed412114 but still requires manual download and decompression.

In the future it would be better to have a `loadMNIST` proc that returns a tuple X_train, y_train, X_test, y_test (or img_train, label_train, ...)

This requires:
  - Downloading the data files from http://yann.lecun.com/exdb/mnist/
  - Uncompressing them:
     - either once this PR lands https://github.com/nim-lang/zip/pull/20
     - or by exposing/reimplementing [`gzReadData`](https://github.com/dom96/untar/blob/master/src/untar/gzip.nim#L58):
        ```Nim
         proc gzReadData(s: Stream, buffer: pointer, bufLen: int): int =
           if bufLen == 0: return 0
           var s = GzStream(s)
           let ret = gzread(s.handle, buffer, bufLen.cint)
           checkZlibError ret, s.handle
           s.pos.inc(ret)
           s.isAtEnd = ret == 0

           return ret
         ```
      - or using the `zip` module [`uncompress`](https://github.com/nim-lang/zip/blob/master/zip/zlib.nim#L274) to string and then read from the StringStream. This is last resort as it is 2 passes, and uncompress requires knowing the length in advance for decompression.

metasyn:
Fixes #168 

Let me know if you'd like things changed, of course :)


src/datasets/mnist.nim
```diff
@@ -95,19 +97,97 @@ proc read_mnist_labels*(labelsPath: string): Tensor[uint8] {.noInit.}=
...
+proc get_mnist_files(): seq[string] =
+  ## Produces the file names for the MNIST dataset
+  ## Returns:
+  ## - seq[string] of filenames
+  return @[
+    "train-images-idx3-ubyte.gz",
+    "train-labels-idx1-ubyte.gz",
+    "t10k-images-idx3-ubyte.gz",
+    "t10k-labels-idx1-ubyte.gz",
+  ]
```
mratsim: This should be a const array

src/datasets/mnist.nim
```diff
@@ -95,19 +97,97 @@ proc read_mnist_labels*(labelsPath: string): Tensor[uint8] {.noInit.}=
...
+
+proc download_mnist_files(files: seq[string]) =
```
mratsim: No input needed, just use the const array

src/datasets/mnist.nim
```diff
@@ -49,31 +49,33 @@
...
-proc read_mnist_images*(imgsPath: string): Tensor[uint8] {.noInit.}=
-  ## Load MNIST images into a Tensor[uint8]
+type mnist = tuple[
+  train_images: Tensor[uint8],
+  test_images: Tensor[uint8],
+  train_labels: Tensor[uint8],
+  test_labels: Tensor[uint8],
+]
+
+const tmp = getTempDir()
```
mratsim: I prefer "global" const in PascalCase to easily spot them.

Non-blocking: regarding datasets, they should probably be put in `.cache/arraymancer/` or something like that, not sure what PyTorch and Keras do. That can be changed later.

This will be important for pretrained models because they weight in hundreds of MB.

src/datasets/mnist.nim
```diff
@@ -49,31 +49,33 @@
...
+
+proc read_mnist_images*(stream: Stream): Tensor[uint8] {.noInit.}=
```
mratsim: No need to export this.

src/datasets/mnist.nim
```diff
@@ -95,19 +97,97 @@ proc read_mnist_labels*(labelsPath: string): Tensor[uint8] {.noInit.}=
...
   ##
   ## MNIST data can be downloaded here: http://yann.lecun.com/exdb/mnist/
-  ## It must be uncompressed before use.  Download and decompression will be automated in the future.
-  # Compressed reader pending https://github.com/nim-lang/zip/pull/20
-  if not existsFile(labelsPath):
-    raise newException(IOError, "MNIST labels file \"" & labelsPath & "\" does not exist")
+  if not existsFile(imgsPath):
+    raise newException(IOError, "MNIST images file \"" & imgsPath & "\" does not exist")
+
+  let stream = newGzFileStream(imgsPath, mode = fmRead)
```
mratsim: Does newGzFileStream also works for uncompressed files?

If no I'd like to have a `if uncompressed`

src/datasets/mnist.nim
```diff
@@ -95,19 +97,97 @@ proc read_mnist_labels*(labelsPath: string): Tensor[uint8] {.noInit.}=
...
+    mnist_domain = "http://yann.lecun.com/exdb/mnist/"
+    client = newHttpClient()
+
+  for f in files:
+    let url = fmt"{mnist_domain}{f}"
+    client.downloadFile(url, tmp & f)
+
+proc delete_mnist_files(files: seq[string]) =
+  ## Deletes the downloaded MNIST files.
+  for f in files:
```
mratsim: just use a const array here as well

src/datasets/mnist.nim
```diff
@@ -95,19 +97,97 @@ proc read_mnist_labels*(labelsPath: string): Tensor[uint8] {.noInit.}=
...
+
+  # Training
+  result.train_images = read_mnist_images(tmp_files[0])
+  result.train_labels = read_mnist_labels(tmp_files[1])
+
+  # Testing
+  result.test_images = read_mnist_images(tmp_files[2])
+  result.test_labels = read_mnist_labels(tmp_files[3])
+
+  delete_mnist_files(tmp_files)
```
mratsim: MNIST is small but I guess for heavy datasets we should provide a way to save permanently and to manage the cached datasets and ML models

tests/datasets/test_mnist.nim
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2018 the Arraymancer contributors
+# Distributed under the Apache v2 License (license terms are at http://www.apache.org/licenses/LICENSE-2.0).
+
+import ../../src/arraymancer
+import unittest
+
+suite "Datasets - MNIST":
+  test "Load MNIST":
+    let mnist = load_mnist()
```
mratsim: mmmh, I have to figure out a way to cache that for CI, I don't think it's cool to download that every commits.

metasyn: I changed this to cache by default in `.cache/arraymancer/` - I also added `.cache` to appveyor

src/datasets/mnist.nim
metasyn: I think it assumes its compressed:

https://github.com/nim-lang/zip/blob/master/zip/gzipfiles.nim#L61

src/datasets/mnist.nim
metasyn: Yeah, that would be cool

Issue #293, opened 2018-09-28
[Autograd] Support unlimited parent tensors


mratsim:
## Context

Currently autograd has a max number of tensors that can be backpropagated:

https://github.com/mratsim/Arraymancer/blob/90263efabdf706d7a74c1d847b7449495d67c124/src/autograd/ag_data_structure.nim#L17

This is limiting for sequence and timeseries prediction as we often split/chunk input into subsequences that are then stacked or concatenated after going through a RNN or a full model.

## Caveats

Unfortunately using a seq would imply that the forward pass will generate 1 heap allocation per layer and an extra pointer indirection as we would have BackpropSeq -> Variable -> Tensor backend. There would be more cache misses and deep learning already stresses memory a lot.

## Solution

Similar to LLVM [SmallVector](http://llvm.org/doxygen/classllvm_1_1SmallVector.html), we can have a type with main data on the stack that extends on the heap via a classic seq when needed.

Regarding stack size in uint64, it probably should be:
  - 6 for data
  - 1 for total_length
  - 1 for the extended backend seq assuming Nim seqs length and reserved are on the heap (not true on v2).
     - [Nim Seq v1 implementation](https://github.com/nim-lang/Nim/blob/c240806756579c3375b1a79e1e65c40087a52ac5/lib/system.nim#L434-L444)
     - [Nim Seq v2 with destructors](https://github.com/nim-lang/Nim/blob/c240806756579c3375b1a79e1e65c40087a52ac5/lib/core/seqs.nim#L15-L23)

This because CPU cache lines are 64 bytes and so can load 8 uint64/ptr per load ([4 cycles on Haswell](https://www.7-cpu.com/cpu/Haswell.html))

## Note regarding MetadataArray

https://github.com/mratsim/Arraymancer/blob/90263efabdf706d7a74c1d847b7449495d67c124/src/tensor/backend/metadataArray.nim#L17-L22

MetadataArray should probably not be replaced by SmallVector as we absolutely don't want to allocate within an OpenMP section, for example due to slicing.

PR #303, opened 2018-09-30
Backpropagation: support unlimited parent tensors

mratsim:
Partially address #293 
This is key for using the stack ops

Baseline benchmark - https://github.com/mratsim/Arraymancer/commit/70dfd2ecdce5bbc996524aa0400f2f6b2a343aea

Note only GEMM calls (including convolution) are parallel as they use BLAS.

### DIFF
--- a/benchmarks/ex02_mnist.nim
+++ b/benchmarks/ex02_mnist.nim
@@ -1,0 +1,61 @@
+import ../src/arraymancer, random
+
+# Make the results reproducible by initializing a random seed
+randomize(42)
+
+let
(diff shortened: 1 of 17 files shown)


mratsim:
Bench of the new scheme. This adds 0.1 min (= 6 seconds) to a 3 min run.



As mentionned in #293, next step is switching to something similar to LLVM small vectors to optimize cache locality.

In the mean time this is reasonable.

mratsim:
Additional xor benchmark, this adds 0.6 second to a 7.5s run so impact is 8%.

Notice the time spent in NimNewSeqOfCap (1.13s)

Before:



After:

Issue #87, opened 2017-10-05
Introduce Scalar[T] concept with rank, shape, strides, offset defined


mratsim:
Scalar types like float32 should be considered tensors of rank 0.

It would be very useful for backpropagation (loss function Scalar -> Tensor)

PR #304, opened 2018-09-30
Make GRU usable in the NN DSL

mratsim:


mratsim:
There is a ugly workaround here:

https://github.com/mratsim/Arraymancer/blob/0ac1cfd2b4d90aeb3f4386f18d45a3916c214c8e/src/autograd/gates_shapeshifting_views.nim#L41-L46

It would be less ugly if Tensor of rank 0 were supported https://github.com/mratsim/Arraymancer/issues/87 or by introducing a SliceGate which backpropagate properly.

Currently we lose track of the tensor, workaround would be to unsqueeze, slice, squeeze.

PR #318, opened 2018-11-20
Replace sugar with future module

bitsnaps:
Since future module has been deprecated, we can use sugar instead.


mratsim:
Thanks!

PR #319, opened 2018-11-20
Some real to float

Vindaar:
Well, here we go again. In https://github.com/nim-lang/Nim/commit/3f6168b3372013c0d4b938e6a83849199d9756c9 `SomeReal` was removed (for good now I suppose).


mratsim:
Perfect

Issue #331, opened 2018-12-11
Implement operations on variables


andreaferretti:
I hope I am not bothering you too much by asking questions in the issue tracker, in case let me know.

This time I am trying to port operations from the tensor level to the variable level. For instance, here is what happens if I try to take the mean along a particular axis for a variable:

```python
# part of python code I am porting
mean = input.mean(-1, keepdim=True)
```

```nim
# trying to do the same in Nim by explicitly passing
# the last axis instead of -1, followed by unsqueeze
template dimensions[TT](v: Variable[TT]): int =
  v.value.shape.len

proc forward(ln: LayerNorm, input: Variable): Variable =
  let
    L = dimensions(input) - 1
    mean = input.mean(L).unsqueeze(L)
```

I get an error

```
Error: type mismatch: got <Variable[Tensor[system.float32]], int>
but expected one of: 
proc mean[TT](a: Variable[TT]): Variable[TT]
proc mean[T: SomeInteger](t: Tensor[T]; axis: int): Tensor[T]
  first type mismatch at position: 1
  required type: Tensor[mean.T]
  but expression 'input' is of type: Variable[Tensor[system.float32]]
proc mean[T: SomeFloat](t: Tensor[T]): T
  first type mismatch at position: 1
  required type: Tensor[mean.T]
  but expression 'input' is of type: Variable[Tensor[system.float32]]
proc mean[T: SomeFloat](t: Tensor[T]; axis: int): Tensor[T]
  first type mismatch at position: 1
  required type: Tensor[mean.T]
  but expression 'input' is of type: Variable[Tensor[system.float32]]
proc mean[T: SomeInteger](t: Tensor[T]): T
  first type mismatch at position: 1
  required type: Tensor[mean.T]
  but expression 'input' is of type: Variable[Tensor[system.float32]]
```

Apparently, the operation of taking the mean of a tensor along some axis **is** already supported. I am not sure how to lift this to the variable level.

I have a similar issue in trying to compute a softmax: it is implemented for tensors, but not for variables. (actually I would need log_softmax, but that's for another moment, let's start with what we already have) Should I define manually a backward propagation operator? Or can autograd help me in these simple cases?

I guess these questions are a little silly, but not being familiar with the inner working of Arraymancer even small issues like these can leave me unsure how to proceed

mratsim:
I don't think those questions are silly at all. The autograd part being still in flux is why I mention in the readme:

> ## Arraymancer as a Deep Learning library
[...]

Issue #301, opened 2018-09-29
Optimize inference


mratsim:
This is the current structure of an autodifferentiable layer:

https://github.com/mratsim/Arraymancer/blob/90263efabdf706d7a74c1d847b7449495d67c124/src/autograd/gates_basic.nim#L22-L60

Only the last part "Caching for backprop" that initializes the result gradient with a zero-ed tensor is skipped in inference mode.

We could skip every `node` operations and even the `gate` part can be skipped. A gate is needed for forward but we could use an `inference` proc instead (so we would have forward, backward and inference), that would avoid several heap allocations.

Issue #227, opened 2018-05-09
Neural Net DSL overhead


mratsim:
#214 introduced a domain specific language for neural net.

It seems like it has a noticeable overhead, at least on a simple perceptron.

benchmark on XOR https://github.com/mratsim/Arraymancer/commit/c19b2050622f5d1ef1ec9a04e1e34f041dc7a59c

with Nim devel from 2018-05-06 (commit hash not displayed for some reason :/)

![2018-05-09_22-19-12](https://user-images.githubusercontent.com/22738317/39837888-fc784c7e-53d7-11e8-8a8f-2194c2d5f38c.png)
![2018-05-09_22-20-53](https://user-images.githubusercontent.com/22738317/39837891-fea56d10-53d7-11e8-81a3-51fe699a033f.png)

mratsim:
False alarm, there is no overhead. The benchmarks are not equivalent because the new syntax automatically add a bias for each linear layer that accounts for the speed difference.

![2018-05-09_22-29-50](https://user-images.githubusercontent.com/22738317/39838151-c9fb05a6-53d8-11e8-9552-f739ee199ee3.png)

--> see the sum line
![2018-05-09_22-28-17](https://user-images.githubusercontent.com/22738317/39838164-d18aa272-53d8-11e8-8af6-4dfd6e23d4d9.png)

PR #333, opened 2018-12-15
Autograd refactor

mratsim:
Refactor autograd to solve #301 and #301

#331 # First pass

The first pass removes methods completely. However this is slower than the previous implementation on the XOR benchmark.

Bench | Method | Proc handler
----|----|----
xor_nodsl (with bias #227)| 7.26~7.45s | 8.06~8.10s
xor_dsl | 9.32~10.06 | 10.36~10.42s

One explanation would be that the method dispatch tree is small so branch prediction is good.

## Second pass
The second pass will optimise inference by not allocating gates during it.


mratsim:
Diving into perf counters it seems like `newSeq` is 1.5~2x more costly after the refactor:

(The global time is +-1s , the current different of 2 s is an extreme case, relative cost of the various function is what is interesting)

PR #351, opened 2019-02-22
Fix image shape ordering

xcokazaki:
When I tried to write image by io_image, I encountered the output image was not what I expected.
What I did was to read image and convert to Tensor without io_image, then write that Tensor with io_image.
My test code is not enough to check the output, however it will write image and can confirm what is happening.

Please confirm the difference the original output and my modification.
- original
test_img_read_write
- modified
test_img_read_write

Please let me know your opinion.


tests/io/test_image.nim
```diff
@@ -26,3 +27,21 @@ suite "[IO] Reading and writing images":
     img.write_png(outpng)
     img.write_jpg(outjpg, quality = 70)
 
+  test "[IO] Reading and Wriging images are equal":
```
Clyybber: Typo in "Wriging"

mratsim:
Good catch, it's a bit annoying for CI that nimblas doesn't have a fallback for stable :/ cc @andreaferretti. I don't think there would be anything broken anyway.

andreaferretti:
What do you mean by a fallback for stable? It's a little too cryptic :-)

Issue #316, opened 2018-11-05
Add an English NLP tokenizer


mratsim:
There is no tokenizer currently which makes parsing text and using word embeddings very hard. I.e. currently it's either a split on white spaces or regexp.

metasyn:
Splitting on white space and regexp seems like it could still be ok?

 - sklearn does the same but provides the ability for the user to give a callable
 https://github.com/scikit-learn/scikit-learn/blob/bac89c2/sklearn/feature_extraction/text.py#L260

 - spacy has (pre/suf/in)-fix regexp in addition, as well as full word regexp:
  https://github.com/explosion/spacy/blob/master/spacy/tokenizer.pyx#L20

In contrast:
- open nlp has a learnable tokenizer, using maximum entropy:
 https://opennlp.apache.org/docs/1.5.3/manual/opennlp.html#tools.tokenizer.introduction

- https://github.com/jirkamarsik/trainable-tokenizer also has a maximum entropy mode

Are you think more of the former? Or is there missing more about how whitespace/regexp tokenization makes the embeddings difficult?

mratsim:
Whitespace is OK for a start but it doesn't work with characters like `?` or `!` or `'`.

Furthermore, having a tokenizer type/API is key for languages that don't use whitespaces at all (like Chinese).

Anyway the first thing is to have a white space tokenizer ;)

PR #352, opened 2019-03-02
Add a naive whitespace tokenizer.

metasyn:
Just beginning #316  - super naive implementation here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ matrix:
     # On OSX we only test against clang (gcc is mapped to clang by default)
     # Note: for OpenMP, Homebrew will build flame/blis with GCC-5
     - os: osx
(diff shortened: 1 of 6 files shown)


metasyn:
Realized that a few of the optional builds were failing - but I've updated travis config to make them pass:

https://travis-ci.org/mratsim/Arraymancer/builds/500940958

Notable changes:
* peg `nimblas` at `0.2.1` instead of `>= 0.2.1` because `0.2.2` now requires nimlang `0.19.9` which is devel right now, not stable.
* the blis bottle that was being downloaded was version 0.2.1, which is quite old, and further more, was hanging on the `make test` step and travis was cancelling the build. I added an if statement to download and build blis from source if we're on osx.

So all builds are passing now 🎉

PR #369, opened 2019-07-22
Add SGD with Momentum Optimizer, addresses #298

### LINKED ISSUE #298: Add backpropagation optimizers
Currently only stochastic gradient descent is supported, at the very minimum it would be nice to support:

- [ ] RMSProp
- [ ] Adam
- [ ] SGD with Momentum
- [ ] SGD with Nesterov Momentum

dylanagreen:
This PR adds momentum with a new SGDMomentum object. The default values for SGDMomentum are the same as SGD, and when left as default the SGDMomentum optimizer will operate in the same way the base SGD optimizer. SGDMomentum optimizers must be declared with `var`, similarly to Adam and unlike SGD. Additionally an SGDMomentum object has a parameter that allows it to use Nesterov momentum rather than regular momentum. I've included a link to the paper that proposes Nesterov momentum in the documentation.

This implementation of SGD with Momentum is within 3-4 decimal places with the PyTorch implementation in both the Nesterov and base momentum cases. I have included a PyTorch output for three SGD test cases (without momentum, with momentum, and with nesterov momentum) as well as one for Adam. Tests for SGDMomentum are located in tests/nn/test_optimizers.nim. The included PyTorch implementation is a modification of the following legacy PyTorch test: https://github.com/pytorch/pytorch/blob/master/test/optim/test.py. 

It may be possible to automate these tests, right now I've done them manually by eye. If the output of the PyTorch impl. is saved to a file the Arraymancer test could load the results and check that at each optimizer step the new values of the (x, y) model points is "close enough" to the PyTorch output.

### DIFF
--- a/src/nn/optimizers/optimizers.nim
+++ b/src/nn/optimizers/optimizers.nim
@@ -28,9 +28,32 @@ type
     ## Stochastic gradient descent
     params*: seq[Variable[TT]]
     lr*: TT.T # Learning rate. T is the generic parameter of Tensor[T]
+    momentum*: TT.T
+    moments: seq[TT]          # Moments for momentum
 
 proc newSGD*[T](params: varargs[Variable[Tensor[T]]], learning_rate: T): SGD[Tensor[T]] {.deprecated: "Use the optimizer macro instead".}=
-  SGD[Tensor[T]](params: @params, lr: learning_rate)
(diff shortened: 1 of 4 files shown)


mratsim:
Perfect, thank you very much

Issue #371, opened 2019-08-25
stdlib_times does not compile with Nvidia NVCC compiler


mratsim:
NVCC may define nanosecond and this seems to conflict with Nim times.

The error message when compiling tests_cuda.nim with just test_init_cuda enabled is

https://github.com/mratsim/Arraymancer/blob/2e7b193ee0554cbbca510a18be16f5564671803c/tests/tensor/test_init_cuda.nim#L15-L33

```
Error: execution of an external compiler program '/opt/cuda/bin/nvcc -c -w -arch=sm_61 --x cu -Xcompiler -fpermissive  -I/home/beta/.choosenim/toolchains/nim-0.20.0/lib -I/opt/cuda/include -I/opt/cuda/include -I/home/beta/Programming/Nim/Arraymancer/tests -o /home/beta/.cache/nim/tests_cuda_d/stdlib_times.nim.cpp.o /home/beta/.cache/nim/tests_cuda_d/stdlib_times.nim.cpp' failed with exit code: 1

/home/beta/.cache/nim/tests_cuda_d/stdlib_times.nim.cpp:322:118: error: expected ‘,’ or ‘...’ before numeric constant
 N_LIB_PRIVATE N_NIMCALL(tyTuple_pfrdooyxJsC6cg34ebuBag, getLocalOffsetAndDst_wPlTAoiD9bfaYQuVXohuphA)(NI64 unix);
                                                                                                                      ^   
/home/beta/.cache/nim/tests_cuda_d/stdlib_times.nim.cpp:326:113: error: expected ‘,’ or ‘...’ before numeric constant
 N_LIB_PRIVATE N_NIMCALL(tyObject_Time_3y2ZpqsTJLqdZvpC9a0rU2Q, initTime_JIXNH9cQMG3SbZ0hXqEYDvA)(NI64 unix, NI nanosecond);
                                                                                                                 ^~~~
/home/beta/.cache/nim/tests_cuda_d/stdlib_times.nim.cpp:1501:118: error: expected ‘,’ or ‘...’ before numeric constant
 N_LIB_PRIVATE N_NIMCALL(tyTuple_pfrdooyxJsC6cg34ebuBag, getLocalOffsetAndDst_wPlTAoiD9bfaYQuVXohuphA)(NI64 unix) {
                                                                                                                      ^   
[...]

PR #372, opened 2019-08-25
Fix cuda/cudnn compilation

mratsim:
This is a potential fix to Cuda compilation being broken with the Nim times module (#371)

This seems to happen because Nvidia compiler NVCC defines (?) `nanosecond` symbol globally which conflicts with the std/times module.

Unfortunately the fix has an annoying version requirements as LLVM releases lag behind Cuda SDK and requires compiling from source to support Cuda 10.1 for example:

| clang++ | supported CUDA release | supported SMs |
| ------- | ---------------------- | ------------- |
| 3.9-5.0 | 7.0-8.0                | 2.0-(5.0)6.0  |
| 6.0     | [7.0-9.0](https://github.com/llvm-mirror/clang/blob/release_60/include/clang/Basic/Cuda.h) | [(2.0)3.0-7.0](https://github.com/llvm-mirror/clang/blob/release_60/lib/Basic/Targets/NVPTX.cpp#L163-L188) |
| 7.0     | [7.0-9.2](https://github.com/llvm-mirror/clang/blob/release_70/include/clang/Basic/Cuda.h) | [(2.0)3.0-7.2](https://github.com/llvm-mirror/clang/blob/release_70/lib/Basic/Targets/NVPTX.cpp#L196-L223) |
| 8.0     | [7.0-10.0](https://github.com/llvm-mirror/clang/blob/release_80/include/clang/Basic/Cuda.h) | [(2.0)3.0-7.5](https://github.com/llvm-mirror/clang/blob/release_70/lib/Basic/Targets/NVPTX.cpp#L199-L228) |
| trunk   | [7.0-10.1](https://github.com/llvm-mirror/clang/blob/master/include/clang/Basic/Cuda.h) | [(2.0)3.0-7.5](https://github.com/llvm-mirror/clang/blob/master/lib/Basic/Targets/NVPTX.cpp#L200-L229) |

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Reminder of supported compilation flags:
 - `-d:blas=mkl`: Use MKL, implies `openmp`
 - `-d:blas=openblas`: Use OpenBLAS
 - by default Arraymancer will try to use your default `blas.so/blas.dll`
-  Archlinux user may have to specify `-d:blas=cblas`.
+  Archlinux users may have to specify `-d:blas=cblas`.
   See [nimblas](https://github.com/unicredit/nimblas) for further configuration.
 - `-d:cuda`: Build with Cuda support
 - `-d:cudnn`: Build with CuDNN support, implies `cuda`.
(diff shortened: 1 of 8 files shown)


mratsim:
Alternative fix pushed upstream https://github.com/nim-lang/Nim/pull/12041

Issue #356, opened 2019-05-31
pca components is incorrect shape


brentp:
The docstring for pca seems to be what I expect, that given input shape of: `[Nb of observations, Nb of features]`, the shape of components should be: `[Nb of observations, Nb of components]`
however, this is not the case in the actual implementation.
 
here is code to reproduce:
```nim
import arraymancer

var x = newSeq[seq[float32]](100)
for r in x.mitems:
  r = newSeq[float32](32)

var xt = x.toTensor
echo xt.shape

var pr = xt.pca(2)

echo pr.results.shape
echo pr.components.shape

#doAssert pr.results.shape[0] == xt.shape[0]
# this fails because components.shape[0] matches xt.shape[1] 
doAssert pr.components.shape[0] == xt.shape[0]
```

Vindaar:
I've never actually worked with PCA, but I have the impression the docstring may simply be wrong?

I fired up Python and checked an example with scikit-learn:
```python
import numpy as np
from sklearn.decomposition import PCA

x = np.array([[90, 60, 90],
              [90, 90, 30],
              [60, 60, 60],
              [60, 60, 90],
              [30, 30, 30]])
print(np.shape(x))
# (5, 3)
pca = PCA(n_components=2)
pca.fit(x)
y = pca.transform(x)
print(y) # equivalent to `pca.results`
# [[-34.37098481 -13.66927088]
#  [ -9.98345733  47.68820559]
#  [  3.93481353  -2.31599277]
#  [-14.69691716 -25.24923474]
#  [ 55.11654576  -6.45370719]]
print(np.shape(y))
# (5, 2)
print(pca.components_) # equivalent to `pca.components`
# [[-0.65580225 -0.4291978  -0.62105769]
#  [ 0.3859988   0.51636642 -0.7644414 ]]
print(np.shape(pca.components_))
# (2, 3)
```

and the equivalent in arraymancer:
```nim
import arraymancer

let xt = [[90'f64, 60, 90],
          [90'f64, 90, 30],
          [60'f64, 60, 60],
          [60'f64, 60, 90],
          [30'f64, 30, 30]].toTensor
echo xt.shape
# [5, 3]
var pr = xt.pca(2)

echo pr.results
# Tensor[system.float64] of shape [5, 2] of type "float64" on backend "Cpu"
# |-34.37098480726787     13.66927088271195|
# |-9.983457328239837     -47.68820559298736|
# |3.934813529880883      2.315992772328602|
[...]

PR #384, opened 2019-09-28
Randomized pca [Ready]

mratsim:
Randomized SVD is implemented.

Several implementations of randomized PCA compatible with fast randomized SVD are available in the benchmark folder cc @brentp:

https://github.com/mratsim/Arraymancer/blob/1d0eb60aae53d9a33bbcc1d36d2f28b4f63a9f12/benchmarks/implementation/pca.nim#L74-L91

### TODO:
- automatically transpose the SVD depending of the matrix shape (i.e for a M by N, is M>N or not)
  this probably should recover the speed difference vs Sikit-learn and fbpca
- update description of pca  #356 see https://github.com/mratsim/Arraymancer/blob/b47b040968268a9a4ed7f57e959d327c2432aeac/src/ml/dimensionality_reduction/pca.nim#L9-L30 
  within the code I do mention `[Nb_obs, Nb_feats] * [Nb_feats, Nb_components]`
- have a pca wrapper around randomized SVD. I think I will probably automatically choose the SVD algorithm depending on the problem size
- breaking: I don't really like the named in result tuple of PCA, I'm thinking of `projected, rotation` a bit similar to R. comments welcome.


brentp:
this is awesome! I have matrices on the order of e.g. 2500x18000 to be reduced to 2500x5 . I can give this PR a try on Monday.
Or, here is some float32 data to test on (Arraymancer seems to required it to be unzipped).
[thousandG.npy.gz](https://github.com/mratsim/Arraymancer/files/3666236/thousandG.npy.gz)

that is data for the 2504 thousand genomes samples at ~18,000 genomic sites. So after projecting to 2504x$n PCS, there should be nice clusters by ancestry that looks like (an unlabeled version of) this:
1kg

mratsim:
Adding gzipped .npy support should be trivial.

You can add a switch here:
https://github.com/mratsim/Arraymancer/blob/b47b040968268a9a4ed7f57e959d327c2432aeac/src/io/io_npy.nim#L66-L81

to use newGzFileStream
https://github.com/mratsim/Arraymancer/blob/b47b040968268a9a4ed7f57e959d327c2432aeac/src/datasets/mnist.nim#L92-L108

It might be worth it to add a `CompressionKind` enum that will be used throughout all file reading proc as .csv or .json can also be compressed.

mratsim:
PR ready.

PCA now reuses the fast randomized SVD, same API as before with extra optional arguments.
The code is between 2x to 10x faster than scikit-learn and 4x-6x faster than Facebook's fbpca.

Potential breaking changes:
```
  - pca input "nb_components" has been renamed "n_components".
  - pca output tuple used the names (results, components). It has been renamed to (scores, loadings).
  - A ``pca`` overload that projected a data matrix on already existing principal_axis
    was removed. Simply multiply the data matrix with the loadings instead.
```

brentp:
with this code:
```Nim
import arraymancer
import plotly
import sequtils
import times

var x = read_npy[float32]("thousandG.npy")

proc pca_svd_tXU[T: SomeFloat](X: Tensor[T], n_components = 2): Tensor[T] {.noInit.}=
  let X = X .- X.mean(axis=0)

  let (U, _, _) = svd_randomized(X.transpose, n_power_iters=2, n_oversamples=10)
  result = X * U[_, 0..<n_components]

var t = cpuTime()
let r = x.pca_svd_tXU(3)
echo "time:", cpuTime() - t

var d = Trace[float32](mode: PlotMode.Markers, `type`: PlotType.Scatter)

d.xs = r[_, 0].toSeq
d.ys = r[_, 1].toSeq

Plot[float32](traces: @[d]).show()
```
The PCA finishes in ~ 3 seconds and makes this plot, which looks great:
mancer-pca

brentp:
with a tensor `x` of shape: `[2504, 17382]`. `x.pca(3)` gives the error:
```
/home/brentp/.nimble/pkgs/arraymancer-0.5.2/ml/dimensionality_reduction/pca.nim(47) pca
/home/brentp/.nimble/pkgs/arraymancer-0.5.2/linear_algebra/decomposition_rand.nim(259) svd_randomized
/home/brentp/.nimble/pkgs/arraymancer-0.5.2/tensor/operators_blas_l2l3.nim(96) *
/home/brentp/.nimble/pkgs/arraymancer-0.5.2/tensor/operators_blas_l2l3.nim(64) gemm
/home/brentp/.nimble/pkgs/arraymancer-0.5.2/tensor/private/p_checks.nim(126) check_matmat
Error: unhandled exception: Number of columns in the first matrix: 8, must be the same as the number of rows in the second matrix: 17382 [IndexError]
```
not sure if this is helpful, but I thought I'd report given the

mratsim:
Thanks, will look into it.

While am at adding new features to PCA, I'm considering a couple of additional options:

- Adding a `preprocess` enum to the signature:
```Nim
type FeaturePreprocessing = enum
  Auto
  NoPreprocessing
  MeanCentering
  MinMaxScaling
  StandardScaling
  RobustScaling
  ...
```
- Alternatively pass a scaler object that will be applied per columns (features)
- Alternatively pass a scaler proc that will be applied per columns (features)

Creating a new thread

mratsim:
Should be really finished for now. Now uses 2x less memory than fbpca and sklearn and is consistently faster:

| Problem size | fbpca (MKL) | scikit-learn (MKL) | Arraymancer (MKL) | Arraymancer (OpenBLAS)
| -----------------| ------- | -------------- | ---------------- | ---------------------|
| 20000x4000 -> 40 components | 2.12s | 3.58s | 0.16s | 0.30s |
| 4000x20000 -> 40 components | 0.40s | 0.21s | 0.09s | 0.12s |

brentp:
thanks for this!
The shapes and naming seem better now. For example, I don't need to transpose the components to project.
And it's extremely fast on something that previously would not finish in hours.

PR #391, opened 2019-10-17
fix #390 by copying `gemm` input tensor

### LINKED ISSUE #390: Ax + b modifies operands
This code:
```
import arraymancer

let x = @[4.0].to_tensor
var A = @[1.0].to_tensor.reshape(1,1)
var b = zeros[float](1);

echo b.to_raw_seq
for _ in 0..4:
  discard (A * x) + b
  echo b.to_raw_seq
```

Mutates `b` and has this output:
```
@[0.0]
@[4.0]
@[8.0]
@[12.0]
@[16.0]
@[20.0]
```

I would expect `b` to not change from `A*x + b`? Some other tests:

 - `x + b` — `b` is unchanged.
 - `b + (A*x)` — `b` is unchanged.

Vindaar:
Apparently when the change to reference semantics was made, this line wasn't changed accordingly. 

I'm not sure if `deepCopy` is the desired way here?

### DIFF
--- a/src/tensor/optim_ops_fusion.nim
+++ b/src/tensor/optim_ops_fusion.nim
@@ -29,8 +29,7 @@ import  ../private/[nested_containers, ast_utils],
 proc tensor_multiplyAdd[T](
   A, B: Tensor[T],
   C: Tensor[T]): Tensor[T] =
-
-  result = C
+  result = deepCopy(C)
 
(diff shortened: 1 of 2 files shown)


mratsim:
You should use clone as C must be contiguous for gemm.

It will also saves memory if C is a result of slicing/reshaping

Issue #389, opened 2019-10-16
Linear system solution (LU)


auxym:
I did some digging around and it seems LU decomposition was recently added to Arraymancer (#382).

*Somewhat sidetracking: any reason why the LAPACK LU result is duplicated in two tensors in `lu_permuted`? It seems like an easy memory optimization to avoid this duplication  (each matrix is half zeros). Scipy's wrapper to `DGETRF`, `scipy.linalg.lu_factor`, returns a single array lu as the wrapped function does.*

Ok, back to the main focus: solving a linear system of equations `AX = B` is a very common thing to do, and LU factorization is a classic, widely applicable and decently efficient way to do it. We could wrap `DGETRS`, except:

1. `DGETRS` wants to do the permutation which is already done in arraymancer's `lu_permuted` wrapper.

2. `DGETRS` expects a single LU matrix as returned by `DGETRF`.

Proposals:

1. Merge `L` and `U` back into a single matrix (`L + U`) and pass a no-op array `IPIV` to `DGETRS`. Easiest but least efficient option.

2. Directly call `DTRSM` from LAPACK twice, which is the routine that actually does the forward-backward subs.

3. Write a pure-nim FBS proc(s) equivalent to `DTRSM`. A naive implementation seems rather straightforward, but I'm not sure if `DTRSM` has additional optimization that would involve more complex implementation and have significant benefits. Some investigation needed here.

4. Don't use `lu_permuted` at all, and wrap `DGESV` directly.

Comments?

mratsim:
> _Somewhat sidetracking: any reason why the LAPACK LU result is duplicated in two tensors in lu_permuted? It seems like an easy memory optimization to avoid this duplication (each matrix is half zeros). Scipy's wrapper to DGETRF, scipy.linalg.lu_factor, returns a single array lu as the wrapped function does._

I approximately follow the high-level `scipy.lu` not `scipy.lu_factor`: https://docs.scipy.org/doc/scipy-1.3.0/reference/generated/scipy.linalg.lu.html?highlight=lu#scipy.linalg.lu

For solving equation via LU, I would call directly the underlying getrf wrapper https://github.com/mratsim/Arraymancer/blob/7fb7e9f1b62691221247186149edbc07402cffc1/src/linear_algebra/helpers/decomposition_lapack.nim#L294-L318
and then pass it to a getrs wrapper.

[...]

PR #404, opened 2019-11-16
Add linear equation solve

auxym:
Re: #389 

Add `solve` proc that is meant to have a similar API to scipy's `linalg.solve`. That is, the different matrix structures handled by LAPACK's various *SV driver routines are selected by an argument to
`solve`, `kind`.

Only the general matrix case, `gesv`, is currently implemented.

### DIFF
--- a/src/linear_algebra/helpers/solve_lapack.nim
+++ b/src/linear_algebra/helpers/solve_lapack.nim
@@ -1,0 +1,47 @@
+# Copyright (c) 2019-Present the Arraymancer contributors
+# Distributed under the Apache v2 License (license terms are at http://www.apache.org/licenses/LICENSE-2.0).
+# This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  nimlapack,
+  ./overload,
+  ../../tensor/tensor
+
+# Wrappers for Fortran LAPACK linear equation driver routines *SV
+# Currently only *GESV is wrapped
+# TODO: Implement GBSV, GTSV, POSV, PBSV, PTSV, SYSV
+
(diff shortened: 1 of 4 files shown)


mratsim:
Awesome, thank you!

I wonder about the MatrixKind, so much focus is given to the general matrix routines. Well we can revisit that later

auxym:
> I wonder about the MatrixKind, so much focus is given to the general matrix routines. Well we can revisit that later

Can you clarify? My understanding is that the non-general cases allow some optimizations for specific matrix structures. For example, I remember reading that Cholesky decomposition as about twice faster for solving pos-def matrices, than LU which is used in the general case.

My own questioning was whether to support the banded/tridiagonal cases, which scipy doesn't seem to do.

mratsim:
For example in the Lapack routines after QR decomposition via geqrf you get the matrix Q in factorized Householder form. There are 2 ways to manipulate that factorized matrix assuming you want to do a multiplication afterwards:
- either [ormqr](https://software.intel.com/en-us/mkl-developer-reference-c-ormqr) which directly multiplies the factorized matrix with the other operand
- or [orgqr](https://software.intel.com/en-us/mkl-developer-reference-c-orgqr) to extract the real matrix from the factorized one and use a normal matrix-multiplication afterwards.

The difference was massive, 30x to 100x speed increase between MKL ormqr and MKL gemm depending on the shape of the matrice.

Commit: https://github.com/mratsim/Arraymancer/pull/384/commits/7d005a8dea93d656ab5ef9bc25898bc0c6dfacc1

PR #408, opened 2019-12-21
Reorder error function inputs for common error metrics

### LINKED ISSUE #380: Reorder error functions inputs y_true and y_pred
All errors functions like `accuracy_score`, `relative_error` and `absolute_error` follow the following signature convention

```Nim
proc my_error[T](y_true, y: Tensor[T]): T =
  discard
```

This is the same convention as scikit-learn, see [accuracy_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html).

However this convention hurts composition via the method call syntax.
For example, inverting the arguments would allow.
```
my_tensor.foo.bar.baz.accuracy_score(true_labels)
```
While currently we are "forced" to use function call syntax:
```
accuracy_score(true_labels, my_tensor.foo.bar.baz)
```

The potential breakage is minimal as all basic error functions are commutative. And it will not break named parameters calls.

Complex loss functions like `softmax_cross_entropy`, `sigmoid_cross_entropy` or `mean_loss` already use the (input, target) ordering.

berquist:
Closes #380 

I'm not sure if I caught all the function signatures, but a `rg -Fl 'y_true'` only hits these files. I also changed a few (but intentionally not most) of the examples using the error functions. Let me know if there is anything else I should change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ let
 for t in 0 ..< 500:
   let
     y_pred = model.forward(x)
-    loss = mse_loss(y_pred, y)
+    loss = y_pred.mse_loss(y)
 
   echo &"Epoch {t}: loss {loss.value[0]}"
 
--- a/benchmarks/ex01_xor.nim
+++ b/benchmarks/ex01_xor.nim
@@ -41,7 +41,7 @@ for epoch in 0..10000:
     # Building the network
     let n1 = relu linear(x, layer_3neurons)
     let n2 = linear(n1, classifier_layer)
-    let loss = sigmoid_cross_entropy(n2, target)
+    let loss = n2.sigmoid_cross_entropy(target)
 
     # Compute the gradient (i.e. contribution of each parameter to the loss)
     loss.backprop()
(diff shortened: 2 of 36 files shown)


mratsim:
Nice thanks, can you add an entry in the changelog?

mratsim:
The examples also should be reordered, for example: https://github.com/mratsim/Arraymancer/blob/f06c43db379d1ef0160315162a45a8ab5bdb2b0d/examples/ex02_handwritten_digits_recognition.nim#L88

To:
```Nim
score += y_pred.accuracy_score(y_test[i*1000 ..< (i+1)*1000]) 
```

berquist:
I did end of changing many instances inside tests, except for:
1. The comments for PCA and SVD
2. Cases where there was some whitespace alignment for `y_pred`:
```
        check:  mean_absolute_error( vec[_, col], expected_vec[_, col]) < 1e-4 or
                mean_absolute_error(-vec[_, col], expected_vec[_, col]) < 1e-4
```
3. Cases where:
```nim
mean_relative_error(expected_grad, expected_sparse_grad)
```

There were also a few leftovers for more complex error functions. If you don't want any of these, we can just nuke those commits.

mratsim:
Thank you!

PR #411, opened 2020-01-09
Enable SIMD on builtin Integer BLAS

mratsim:
This enables SIMD vectorization on the builtin integer BLAS.

It's about 3.5 faster on my machine on reasonably sized matrices (1500x1500).
Note that besides MKL which only works on Intel CPU, no other BLAS vendor (OpenBLAS, Apple Accelerate, BLIS) provide accelerated integer matrix multiplication.

Benchmarks vs Julia and Numpy on AVX512 CPU

Nim 1.0.4. Compilation option: "-d:danger -d:openmp"
Julia v1.3.1
Python 3.8.1 + Numpy-MKL 1.18.0

Nim: 0.14s, 22.7Mb - (~0.4ms single threaded)
Julia: 1.67s, 246.5Mb
Python Numpy: 5.69s, 75.9Mb


metasyn:
🎉

Issue #327, opened 2018-12-09
Custom NN types - "undeclared field 'grad'


mratsim:
When creating a custom NN type if it doesn't use generics you run into issues with the autograd `backward` method:

For example
```Nim
type ShakespeareNet = object
  # Embedding weight = Encoder
  encoder_w: Variable[Tensor[float32]]

  # GRU RNN = Internal representation
  gru_W3s0, gru_W3sN, gru_U3s, gru_bW3s, bU3s: Variable[Tensor[float32]]

  # Linear layer weight = Decoder
  decoder_w: Variable[Tensor[float32]]
```

Will trigger an error here:

https://github.com/mratsim/Arraymancer/blob/89cd3f4cbac7a9219d404c23915bd06da5d1098c/src/autograd/gates_basic.nim#L30-L31

```
yourfile.nim(71, 22) template/generic instantiation of `backward` from here
../src/autograd/gates_basic.nim(32, 34) Error: undeclared field: 'grad'
```

This happens when the generic types and its method can be materialised early.

The workaround is to use a generic type ...

```Nim
type ShakespeareNet[TT] = object
  # Embedding weight = Encoder
  encoder_w: Variable[TT]

  # GRU RNN = Internal representation
  gru_W3s0, gru_W3sN, gru_U3s, gru_bW3s, bU3s: Variable[TT]

  # Linear layer weight = Decoder
  decoder_w: Variable[TT]
```

Renaming the payload field from variable to something else to avoid collision might also help.
Removing `methods` altogether would also be a solution (https://github.com/mratsim/Arraymancer/issues/302)

mratsim:
Trying to change the following

```Nim
type
  Sgd*[TT] = object
    ## Stochastic gradient descent
    params: seq[Variable[TT]]
    lr: float32 # Learning rate.
```

into

```Nim
type
  Sgd*[T] = object
    ## Stochastic gradient descent
    params: seq[Variable[Tensor[T]]]
    lr: T # Learning rate.
```

also triggers the same issue at the same location.

Changing `payload.variable` field to a name that is not used at all didn't fix the issue:

https://github.com/mratsim/Arraymancer/blob/3ae364dc31647ac580464979d251868a25629ce5/src/autograd/ag_data_structure.nim#L81-L86

mratsim:
fixed by #333

PR #413, opened 2020-02-11
WIP: Add dbscan clustering method

abieler:
First shot at implementing DBSCAN in nim.
Algorithm from sklearn

- [x] `debug.nim` to be removed
- [x]  `moons.csv` to be removed
- [x]  fix bug for the last record of the data tensor
- [ ] replace brute force nearest neighbor calculation with something smarter
- [ ] tests to be added

Basic usage:
```nim
import dbscan

let X = toTensor(...)
let metric = Euclidean()
let labels = dbscan(X, 0.25, 5, metric)
```

Please weigh in on my preliminary design decision with the `metrics` and the use of multimethods for the `pairwise()` function.

FYI: nim newbie here, so mostly having fun / toying around with the language so far.

Cheers

### DIFF
--- a/src/ml/clustering/dbscan.nim
+++ b/src/ml/clustering/dbscan.nim
@@ -1,0 +1,49 @@
+import deques
+import sequtils
+
+import ../metrics/distance
+import ../neighbors/neighbors
+import ../../tensor/tensor
+
+
+proc findCorePoints(neighborhoods: seq[seq[int]], minSamples: int): seq[bool] =
+  let nSamples = len(neighborhoods)
+  result = repeat(false, nSamples) 
+  for i in 0..high(neighborhoods):
+    if len(neighborhoods[i]) >= minSamples:
+      result[i] = true
+
+
+proc assignLabels(is_core: seq[bool], neighborhoods: seq[seq[int]]): seq[int] =
+  let nSamples = len(neighborhoods)
+  var
+    label_num, j, v: int
+    labels = repeat(-1, nSamples)
+    neighbors: seq[int]
+    stack = initDeque[int]()
+
+  for i in 0..<nSamples:
+    if labels[i] != -1 or is_core[i] == false:
+      continue
+    j = i
+    while true:
+      if labels[j] == -1:
+        labels[j] = label_num
+        if is_core[j]:
+          neighbors = neighborhoods[j]
+          for j in 0..high(neighbors):
+            v = neighbors[j]
+            if labels[v] == -1:
+              stack.addLast(v)
+      if len(stack) == 0:
+        break
+      j = peekLast(stack)
+      popLast(stack)
+    label_num += 1
+  return labels
+
+
+proc dbscan*[T: SomeFloat](X: Tensor[T], eps: float, minSamples: int, metric: PairwiseDist): seq[int] =
+  let neighborhoods = simpleNearestNeighbors(X, eps, metric)
+  let is_core = findCorePoints(neighborhoods, minSamples)
+  result = assignLabels(is_core, neighborhoods)
--- a/src/ml/metrics/distance.nim
+++ b/src/ml/metrics/distance.nim
@@ -1,0 +1,51 @@
+import math
+import sets
+import sequtils
+
+import ../../tensor/tensor
+
+
+type
+  PairwiseDist* = ref object of RootObj
+  Euclidean* = ref object of PairwiseDist
+  Manhattan* = ref object of PairwiseDist
+  Minkowski* = ref object of PairwiseDist
+    p*: int
+  Jaccard* = ref object of PairwiseDist
+
+
+method computeDistance(metric: PairwiseDist, v, w: Tensor[float]): float {.base.} = 
+  return 0.0
+
+
+method computeDistance(metric: Manhattan, v, w: Tensor[float]): float = 
+  return sum(abs(v -. w))
+
+
+method computeDistance(metric: Minkowski, v, w: Tensor[float]): float = 
+  let p = float(metric.p)
+  let d = abs(v -. w)
+  result = pow(sum(mapIt(d, pow(it, p))), 1/p)
+
+
+method computeDistance(metric: Jaccard, v, w: Tensor[float]): float = 
+  let sx = toHashSet(toSeq(v))
+  let sy = toHashSet(toSeq(w))
+  result = len(intersection(sx, sy)) / len(union(sx, sy))
+
+
+method pairwise*(metric: PairwiseDist, x, y: Tensor[float]): Tensor[float] {.base.} =
+  let n_obs = x.shape[1]
+  result = newTensor[float](n_obs, n_obs)
+  for i, v in enumerateAxis(x, 1):
+    for j, w in enumerateAxis(x, 1):
+      if j < i:
+        result[i,j] = result[j,i]
+        continue
+      result[i,j] = computeDistance(metric, v, w)
+ 
+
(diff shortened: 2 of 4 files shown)


mratsim:
Thanks, this looks quite simple and maintainable.

#### Regarding methods:
- you're actually using single dispatch since the only polymorphic part is the `Metric`
- generic methods are in bad shape in Nim and Tensor[float] is generic over float32 and float64:
  - https://forum.nim-lang.org/t/4921
  - https://github.com/nim-lang/Nim/issues/6283
  - I had to remove them from Arraymancer
    - https://github.com/mratsim/Arraymancer/issues/327 -> https://github.com/mratsim/Arraymancer/pull/333
- minor: since method dispatch at runtime, the `pairwise` will always end-up in the final binary even if unused (no dead code elimination).

#### Alternative 1 - enum-dispatch (not advised)

Use an static enum for dispatching, for example replace
```Nim
type
  Metric* = ref object of RootObj
  Euclidean* = ref object of Metric
  Manhattan* = ref object of Metric
  Minkowski* = ref object of Metric
    p: int
  Jaccard* = ref object of Metric


method pairwise*(x, y: Tensor[float], metric: Metric): float {.base.} =
  return -1

method pairwise*(x, y: Tensor[float], metric: Euclidean): float =
  var d: float
  for v, w in zip(x, y):
    d = (v - w)
    result +=  d * d
  result = sqrt(result)


method pairwise*(x, y: Tensor[float], metric: Manhattan): float =
  for v, w in zip(x, y):
        result += abs(v - w)


method pairwise*(x, y: Tensor[float], p:float, metric: Minkowski): float =
    var d: float
    for v, w in zip(x, y):
        d = abs(float(v - w)) 
        result += pow(d, p)
    result = pow(result, 1/p)
```

by

```Nim
type
  Metric* = enum
    Euclidean
    Manhattan
    Minkowski
    Jaccard

proc pairwiseEuclidean(x, y: Tensor[float]): float =
  var d: float
  for v, w in zip(x, y):
    d = (v - w)
    result +=  d * d
  result = sqrt(result)


proc pairwiseManhattan(x, y: Tensor[float]): float =
  for v, w in zip(x, y):
    result += abs(v - w)


proc pairwiseMinkowski(x, y: Tensor[float], p:float): float =
  var d: float
  for v, w in zip(x, y):
      d = abs(float(v - w)) 
      result += pow(d, p)
  result = pow(result, 1/p)

proc pairwise[T](x, y: Tensor[T], metric: static Metric, p = T(NaN)): float =
  when metric == Euclidean:
    pairwiseEuclidean(x, y)
  elif metric == Manhattan:
    pairwiseManhattan(x, y)
  elif metric == Minkowski:
    pairwiseMinkowski(x, y, p)
  else:
    {.error: "Unreachable".}
```

This works because I expect the enum to not grow wildly in the future hence we don't suffer much from the Expression Problem which in short is:
- adding new types to enum-based dispatch is hard because you need to update all **existing** functions while adding new functions is easy, you should dispatch on all existing types and don't need to update existing code.
- adding new functions/interface to method-based dispatch is hard because you need to create one for each **existing** types (even with NotImplementedError) while adding new type is easy, create the type and all relevant methods

i.e. if you need to add a new function/types, can you do it in a self-contained extra file or do you need to edit existing code or reformulated: is it easy to add operations or datatypes.

#### Alternative 2 - typedesc dispatch (advised)

Nim offers almost dependent-types and they can be used for compile-time dispatch. They are IMO much nicer to work with than static enums in many-cases.

```Nim
type
  Euclidean* = object
  Manhattan* = object
  Minkowski* = object
  Jaccard* = object

proc pairwise(metric: typedesc[Euclidean], x, y: Tensor[float]): float =
  var d: float
  for v, w in zip(x, y):
    d = (v - w)
    result +=  d * d
  result = sqrt(result)

proc pairwise(metric: typedesc[Manhattan], x, y: Tensor[float]): float =
  for v, w in zip(x, y):
    result += abs(v - w)

proc pairwise(metric: typedesc[Minkowski], x, y: Tensor[float], p:float): float =
  var d: float
  for v, w in zip(x, y):
      d = abs(float(v - w)) 
      result += pow(d, p)
  result = pow(result, 1/p)
```

You get the best of both world:
- Adding a new type, doen't require to change a when/if/case statement in all existing function
- Adding a new function to a datatype doesn't require you to add one for each types

As a side-effect, this doesn't use the name `Metric`, there might be more obvious use of this name so keeping this option open for later is good.

abieler:
thanks for your valuable input! as i am trying out the `typedesc` version i noticed that
before i would be able to do:

```nim
let metric = Euclidean()
let X = toTensor(....)
var result = dbscan(X, ...., metric)
# or
let metric = Minkowski(p=4)
var result = dbscan(X, ...., metric)
```

so the api for `dbscan()` remains the same, independent of the number of parameters of the metric.
Now with `typedesc` I think I would have to do

```nim
let X = toTensor(....)
var result = dbscan(X, ...., Minkowski, p=4)
# or
var result = dbscan(X, ...., Manhattan)
```
hence come up with two implementations of `dbscan()`. am I missing something here?


cheers
andre

src/ml/cluster/dbscan.nim
```diff
@@ -0,0 +1,78 @@
+import algorithm
+import deques
+import sequtils
+
+import ../metrics/distance
+import ../../tensor/tensor
+
+
+proc argsort[T: SomeFloat](a: seq[T]): seq[int] =
+  result = toSeq(0..<high(a))
```
brentp: should this be `toSeq(0..high(a))` ?

mratsim: yes or `toSeq(0..<a.len)`

mratsim:
You can have:

```Nim
type
  Euclidean* = object
  Manhattan* = object
  Minkowski* = object
  Jaccard* = object

proc dbscan*[T: SomeFloat](X: Tensor[T], eps: float, minSamples: int, metric: typedesc[Euclidean or Manhattan or Jaccard]): seq[int] =
    let neighborhoods = computeNeighborhoods(X, eps, metric)
    let is_core = findCorePoints(neighborhoods, minSamples)
    result = assignLabels(is_core, neighborhoods)

proc dbscan*[T: SomeFloat](X: Tensor[T], eps: float, minSamples: int, metric: typedesc[Minkowski], p: int): seq[int] =
    let neighborhoods = computeNeighborhoods(X, eps, metric, p)
    let is_core = findCorePoints(neighborhoods, minSamples)
    result = assignLabels(is_core, neighborhoods)
```

which seems reasonable in terms of duplication.

Vindaar:
@abieler any chance you'll continue this PR in the near future? I might want to give this a try for my work soonish (meaning in a few weeks probably). If not I'll probably pick up from your PR at that point, if that's ok with you.

abieler:
Hey @Vindaar I ll give it another crack over the Easter break

abieler:
@Vindaar  or anyone feel free to hijack this PR. i unfortunately won't have a lot of time to invest here :(

the code is working and i think ok efficient except for the neighborhood computation which is a naive linear search. replacing it with kdtree or so would definitely be a huge improvement.

mratsim:
Thanks @abieler

Vindaar:
No worries and thanks for your work!

I started a kd tree implementation, which is like half done in arraymancer recently for something else. I might rebase my code on your PR and use try to make use of that.

Vindaar:
With #518 merged, I'll close this.

Thanks again @abieler!

PR #415, opened 2020-02-16
fix typo in julia version

oxinabox:
The latest julia release on nightly is `v1.5.0-DEV` I thus thinl that julia `v6.0.0` was a typo,
and it should have been `v0.6`


mratsim:
you're right!

PR #416, opened 2020-02-18
Update docs

christopherzimmerman:
In the speed section of the docs, there was a typo that stated a runtime as being o(n^4), when it should be o(4*n)


mratsim:
Thanks!

Issue #400, opened 2019-11-04
masked_select: Filtering tensors with a boolean mask tensor


brentp:
given a 2d array/tensor, how would i do the arraymancer equivalent of this numpy expression:
```python
a[:,a.mean(axis=0) > 0.5] = -1
```

I think it's probably some combination of the `map/apply/fold_inline`, but it's not obvious how to do that.
Once I understand, I can open a PR with some common examples like this if it's helpful.

brentp:
I have come up with this:
```
import arraymancer

var T = randomTensor[float32](8, 3, 1'f32)
let m = T.mean(axis=0).broadcast(T.shape)

apply2_inline(T, m, if x > y: x else: -1)
echo T
```
which seems reasonable. not sure if there's a simpler way.

mratsim:
So what's happening is that Numpy accepts a Tensor of booleans as a mask for selecting things.

Your solution works for your case because you don't need to discard/filter the values, however implementing filtering with a Tensor of bool would also be very useful for dataframes/analysis related stuff.

For filtering, I see the following difficulties:
- Implementing the algorithm: it may be easier to use PyTorch's `masked_select` as a reference but while not as complex as Numpy, the codebase is very complex at the moment due to their 3 backens mixed together: C Torch, Aten and C10.
- Updating the indexing macro to accept a Tensor of bool and dispatch to the `masked_select` proc(s). This may require multiple dispatch/allocations if 2 dimensions are sliced at the same time.

brentp:
Thanks for considering it. For my additional 0.02...
I guess that, for example indexing with a tensor of booleans or indexes as in numpy is a fundamental feature, but for other stuff, I think I'd be fine just writing the 2-3 lines myself if I had a better handle on the `map*` and `apply*` functions.

To that end, it'd be nice if the broadcasting was done auto-magically (as in numpy) as well.

mratsim:
Broadcast is done auto-magically with the `.+` and other series of dot function. I intentionally have 2 different operators because it's a pain in Numpy to not broadcast and have a silent error to debug.

brentp:
I am also having trouble just using the map/apply stuff. for example, I can't understand how to change this (which either segfaults or runs endlessly):
```Nim
import arraymancer

var T = randomTensor[float32](250, 17384, 1'f32)
let m = T.map_inline():
  if x < 0: 1'f32 else: 0'f32

echo m.shape
echo m
echo m.mean(axis=0)
```

[...]

PR #429, opened 2020-04-05
[RFC] Masked select and Masked fill

mratsim:
This is the first step in supporting NumPy fancy/advanced indexing.
The second step being extending the `[]` and `[]=` macros to automatically dispatch
to either of the `index_select`, `masked_select`, `masked_fill` or the about 10 other slicing variations that already exist.

Thanks Base2 Genomics for sponsoring the work (https://base2genomics.com/)

## State

More tests to be added and documentation to be improved.
masked_axis_fill with a tensor instead of just a scalar which is very common for dataframes.

The current PR is currently more of a RFC to discuss proc names and clarify documentation.

## API design

Turns out that the API is not straightforward at all and it's very easy to get confused, feedback welcome, that would also significantly help in improving the documentation:

### masked_select

Inspiration Numpy: https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#boolean-array-indexing
image

and PyTorch: https://pytorch.org/docs/stable/torch.html#torch.masked_select

image

In Arraymancer:
https://github.com/mratsim/Arraymancer/blob/7db6023cc00821416ebd9e5938f5ac073978b6d6/tests/tensor/test_selectors.nim#L43-L57

Signature: masked_select(t, mask) -> 1D Tensor

#### Tricky part

This returns a 1D Tensor like Numpy and PyTorch. As shown in this Stack overflow thread: https://stackoverflow.com/questions/51586364/tf-boolean-mask2d-2d-gives-1d-result
some users may want to filter the non-zero from `[[1,0],[2,3]]` and obtain `[[1], [2, 3]]`.
This is not a valid "dense" tensor as [1] does not have the same shape as [2, 3].

Tensorflow provides `RaggedTensor` and PyTorch is exploring NestedTensor (https://www.tensorflow.org/guide/ragged_tensor) for this case.

### masked_axis_select

Inspiration Numpy: https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#boolean-array-indexing
image

and Tensorflow: https://www.tensorflow.org/api_docs/python/tf/boolean_mask
image

#### Tricky part

The mask is 1D and corresponds to axis indices that will be retained or dropped.

Tensorflow allows multi-dimensional mask, I don't know at all what kind of result I should expect from  a 2D masking operation with an axis so I assume it's Python overloading flexibility and combining both will result in an error

image

### masked_fill

Straightforward: take a mutable tensor and a mask of the same shape and if the mask element is true fill with value.

### masked_axis_fill

Inspiration Numpy: https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#boolean-array-indexing 
image
(from https://github.com/mratsim/Arraymancer/issues/400#issue-517264050)

https://github.com/mratsim/Arraymancer/blob/7db6023cc00821416ebd9e5938f5ac073978b6d6/tests/tensor/test_selectors.nim#L91-L102

#### Tricky part

The mask is 1D and corresponds to axis indices that will be filled or skipped.

### masked_filled_along_axis

Inspiration: my flawed initial understanding of Numpy boolean masking

It takes a N-D mask, iterate along the axis of a tensor and apply `masked_fill` on the axis slice.

https://github.com/mratsim/Arraymancer/blob/7db6023cc00821416ebd9e5938f5ac073978b6d6/tests/tensor/test_selectors.nim#L105-L116

#### Tricky part

Notice how `masked_axis_fill` and `masked_filled_along_axis` have very similar names, similar application but it's easy to confuse the axes:

https://github.com/mratsim/Arraymancer/blob/7db6023cc00821416ebd9e5938f5ac073978b6d6/tests/tensor/test_selectors.nim#L96
https://github.com/mratsim/Arraymancer/blob/7db6023cc00821416ebd9e5938f5ac073978b6d6/tests/tensor/test_selectors.nim#L110


brentp:
awesome!
I've never needed mask with > 1 dimension. And those "limitations" seem sane (I also never use ragged arrays).
I should clarify that it's not me personally, it's a [small genomics company](https://base2genomics.com/) I co-founded that's sponsoring the work.

Vindaar:
> Turns out that the API is not straightforward at all and it\'s very
> easy to get confused, feedback welcome, that would also significantly
> help in improving the documentation:

Heh, indeed. I remember learning fancy indexing in numpy the first
time... and by now I forgot the little intricacies again (which is very
telling).

### `masked_select`

I think for `masked_select` we can stress (either by a change of name or
in the documentation) that it should be considered as an action on the
underlying raw data, which is stored in a 1D form anyways.

That way it isn\'t surprising that the shape information is not retained
and data is returned as a 1D tensor.

Supporting returning data in a non 1D form is imo another story and not
necessarily part of a PR that introduces `masked_select`. Unless
arraymancer support for sparse tensors I don\'t think it\'s really a
good fit. Every use case I have, in which I have ragged data is simply
not something I use arraymancer for. That\'s exactly the reason my
[TimepixAnalysis](https://github.com/Vindaar/TimepixAnalysis/tree/master/Analysis/ingrid/)
codebase uses arraymancer sparingly. The data I deal with is extremely
sparse. The Timepix is an ASIC with 256x256 pixels. The events I look at
usually have less than 400 active pixels. Thus, I have many
`seq[seq[T]]` data structures around. There\'s just no nice way of
packaging such data and that\'s life imo. Forcing a tool to work with
something it was never meant to doesn\'t sound like a great idea to me.

Regarding the name, I never quite understood why a proc like this in
numpy etc. is not just called `filter`? That\'s all it really does. It
takes conditions and returns elements that match the condition. I think
people in functional programming land came up with pretty good names for
this sort of stuff.

Sure, `masked_select` can be considered the underlying operation that
would make a `filter` operation work. But given that the usage of such a
proc would usually result from fancy indexing, thus avoiding names in
general, it would imo make sense to give the underlying proc a useful
name that doesn\'t make me feel like the proc is discouraged from use.

The only argument against a name change would be that people coming from
torch, numpy etc. would be confused by the name, imo.

### `masked_axis_select`

Here again I feel like it makes sense that the input is a 1D tensor. To
me the name implies that the given mask works along an axis. An axis is
by definition 1 dimensional, no matter how many dimensions the objects
along that axis have (which of course are N - 1 dimensional for an ND
tensor).

Supporting multiple dimensions as an input to an \"axis select\" seems
weird. I would stick to well defined assumptions that don\'t leave room
for ambiguities. If a need for a multi dimensional selection were to
arise, I\'d give that a different name.

This could also be `filter_axis`.

### `masked_fill`

As you say, pretty straight forward.

Could also take inspiriation from functional programming and call it
`apply`, although in this case I feel like the term `apply` is less
clear than `filter` for `masked_select`.

### `masked_axis_fill`

In the same vein, this could be `apply_axis`.

This makes sense to me for the same reason `masked_axis_select` does.

### `masked_filled_along_axis`

This does indeed confuse me.

Is this not essentially the inverse of the theoretical
`masked_axis_select` for ND tensors?

And due to adding the `axis` argument, we essentially restrict the usage
of the proc.

This is really confusing me. I think the reason is that it breaks my
idea of well defined input -\> output. Since I don\'t know clearly what
the mask argument represents if it\'s the result of a comparison, it
makes it super hard to reason about what the meaning of `axis` is going
to be, because as far as I understand it, axis does not refer to the
tensor acted upon, but rather on the mask.

I don\'t know, maybe I\'m just not thinking straight:

If I understand correctly this does not provide functionality that could
not either be done via `masked_fill` or `masked_axis_fill` already.

One thing that comes to mind, to support:

``` {.nim}
a.masked_fill_along_axis(a.sum(axis = 0) .> 1, axis = 0, -10) 
```

for `masked_axis_fill` instead:

``` {.nim}
a.masked_axis_fill(a.sum(axis = 0) .> 1, axis = 0, -10) 
```

would be to allow ND tensors as arguments to `masked_axis_fill`, but if
and only if the the input represents a squeezable tensor. Just to
provide convenience for results like `a.sum(axis = 0) .> 1`, which (as
far as I can read from your example) do not return an (N-1)D tensor, but
rather an ND tensor with only one element along the reduced axis.

But maybe this could also happen as part of reducing functionality.

mratsim:
PR ready:

- I have added the possibility to fill an axis with a tensor which is a common usecase for dataframes: https://github.com/mratsim/Arraymancer/blob/38f048c44ad364d65ed547b15f61d04d864160ac/tests/tensor/test_selectors.nim#L121-L141
- I have given an example of 2D masking for masked_fill_along_axis. AFAIk it's not possible to do the same with the other proc, if you know a Numpy function that does the same I would love to ~steal~ know the name: https://github.com/mratsim/Arraymancer/blob/38f048c44ad364d65ed547b15f61d04d864160ac/tests/tensor/test_selectors.nim#L157-L209
- Added a missing test for masked_fill

The next step would be to integrate those and index_select in the `[]` and `[]=` macro.

PR #444, opened 2020-04-29
add kernel density estimation (KDE)

Vindaar:
This is a WIP implementation for kernel density estimation for 1D tensors. Functionality wise it's done.

I'm already opening this as a draft, because I have a few questions:
Where should the following procs go:
- `linspace`, `logspace` (latter not needed, but why not add it while I'm at it)
  -> moved to `init_cpu.nim`
- `gauss`
  -> it and the other kernel functions moved to new `stats/distributions.nim`
- `sort`, `sorted`
  -> moved to new `algorithms.nim`
- `percentile`, `iqr`
  -> moved to `aggregate.nim`
?
Also not sure if `stats` fits `kde` itself or if there's a better place for it.

Aside from those questions (and obviously removing the debug code), I need to write a few lines of documentation and add tests for the above procs. Finally, I'll also add the option to include a `weights` tensor to assign a weight to each input value (by default each value in a KDE is assumed to have weight of 1, but when calculating a KDE for e.g. prebinned data assigning weights is useful).

As I said on gitter, the generalization to ND if desired can be added at a later time when looking up nearest neighbors can be done more efficiently (kd-tree?).

edit: Mostly done now. Only remains a test for the KDE and the addition of weights.

edit2: I turned this into a regular PR now. I'll look over it again tomorrow to see if I forgot something.


Vindaar:
From travis: 

> [GC] cannot register global variable; too many global variablesError: execution of an external program failed: '/home/travis/build/mratsim/Arraymancer/build/tests_cpu '

Now that's a fun error I haven't seen so far!

Vindaar:
Sorry for mixing in fixes for fancy indexing here. I stumbled on the issue of the last commit https://github.com/mratsim/Arraymancer/pull/444/commits/43364400c54230d5b6d96d70aa0fd562e8cb90db while writing a test for `argsort`.

Vindaar:
I've implemented weights support for the KDE too now. Stumbled on a rather weird bug of fancy indexing. If I do something like:

```nim
    t = t[sortIdx]
```

the resulting `t` is full of garbage data. I have to explicitly `clone` the `t[sortIdx].clone`  to make it work.

Aside from that I played around with boundary corrections. For most data I reckon the boundary biases don't matter too much, since the bandwidth will be relatively small compared to the total range covered by the input. If that's not the case however, it becomes problematic, since the boundaries will be extremely under represented. But this seems to be something most implementations in other languages don't care about.

I wrote a simple mirror boundary correction by adding mirrored data covering a full bandwidth to either side of the data and weights. It works, but for my use case the result was still wayy worse than applying a reasonable guess based on what my input looks like and extend it manually that way (and in that case it was only a linear dependency). Guess by using a spline interpolation near the boundary to extend it in a reasonable direction would work better, but I think for the time being I'll leave this stuff out.

Vindaar:
So the tests simply fail sometimes, because there are too many globals. 

I see you have those split_tests in the repository, which you used because it used too much RAM in the past. Maybe we have to revive that?

src/stats/kde.nim
```diff
@@ -0,0 +1,209 @@
...
+      (start, stop) = findWindow(cutoff, t[i], x, start, stop)
+      # TODO: rewrite using kernel(t: Tensor) and fancy indexing?
+      for j in start ..< stop:
+        result[j] += norm * kernel(x[j], t[i], bwAct)
+  else:
+    for i in 0 ..< t.size:
+      (start, stop) = findWindow(cutoff, t[i], x, start, stop)
+      # TODO: rewrite using kernel(t: Tensor) and fancy indexing?
+      for j in start ..< stop:
+        result[j] += weights[i] * norm * kernel(x[j], t[i], bwAct)
```
Vindaar, line 177: I tried to use a template:
```nim
  template loop(arg: untyped): untyped =
    for i {.inject.} in 0 ..< t.size:
      (start, stop) = findWindow(cutoff, t[i], x, start, stop)
      for j {.inject.} in start ..< stop:
        arg
```
but it doesn't work. Fails with undeclared identifier `j`, sigh.

Vindaar:
So I guess aside from the following aforementioned problem with fancy indexing and assigning that to the same tensor, I wonder:

- should we implement `sort` and family properly directly? I.e. to handle ND tensors?
- same for `percentile` and `iqr`?

src/stats/kde.nim
```diff
@@ -0,0 +1,209 @@
...
+  proc `fn Kernel`*(x, x_i, bw: float): float {.inline.} =
+    let val = (x - x_i) / bw
+    result = fn(val)
+
+makeKernel(box)
+makeKernel(triangular)
+makeKernel(trigonometric)
+makeKernel(epanechnikov)
+
+# manually build the gauss kerne, since the underlying distribution relies
```
mratsim, line 43: nitpick: manually build the gauss kerne

src/stats/kde.nim
```diff
@@ -0,0 +1,209 @@
...
+type
+  KernelKind* = enum
+    knCustom = "custom"
+    knBox = "box"
+    knTriangular = "triangular"
+    knTrig = "trigonometric"
+    knEpanechnikov = "epanechnikov"
+    knGauss = "gauss"
+
+  KernelFunc* = proc(x, x_i, bw: float): float {.inline.}
```
mratsim, line 31: Does inline actually work for a callback?

src/stats/kde.nim
```diff
@@ -0,0 +1,209 @@
...
+      result[1] = j
+      break
+    inc j
+ # set to max, if we left the while loop naturally
+  if result[1] == 0: result[1] = t.size
+  assert result[1] > result[0]
+
+proc kde*[T: SomeNumber; U: int | Tensor[SomeNumber] | openArray[SomeNumber]](
+    t: Tensor[T],
+    kernel: KernelFunc,
```
mratsim, line 92: should this be `static` if we want `{.inline.}` to work (though I'm pretty sure static proc will crash the compiler).

src/stats/kde.nim
mratsim: What if you use {.dirty.}? Or `mixin j`

I guess the undeclared identifier is due to the template being called in a generic proc

Vindaar:
> Are iqr and kde well known?

**kde**

I may be biased, but to me kde seems to be rather common. 
- scipy: 
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.html#scipy.stats.gaussian_kde
- R: 
https://www.rdocumentation.org/packages/ks/versions/1.10.7/topics/kde

Judging by the wikipedia entry about implementations:
https://en.wikipedia.org/wiki/Kernel_density_estimation#Statistical_implementation
kde and kdensity are both common enough.

Personally I don't really mind. 

**iqr**

Both scipy and R also uses the name iqr:
- https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.iqr.html
- https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/IQR


Given that arraymancer already prefers shorter notation over longer (e.g. std instead of stdlib's standardDeviation) I think it is consistent.

src/stats/kde.nim
Vindaar: Ah, right. Forgot about that "template in generic" gotcha again.

I tried `{.dirty.}`, that didn't help. I'll give `mixin` a try.

src/stats/kde.nim
Vindaar: Oh, good point. I wasn't thinking about callbacks when I defined the type. But it is one of course. I would be surprised if inlining that worked, but I can check the generated code.

src/stats/kde.nim
Vindaar: I'll give it a try.

src/stats/kde.nim
Vindaar: Decided to have the actual `kde` implementation proc to take `static KernelFunc` and clean up the convenience `kde` implementation by using a template. Works well and should allow inlining (at least to "strip" the kernel wrapper of the underlying distribution proc).

mratsim:
Seems like `unittest` globals finally broke

https://travis-ci.org/github/mratsim/Arraymancer/jobs/683938913#L1344

```
[GC] cannot register global variable; too many global variables
```

I'll have a solution in master this evening

mratsim:
Excellent PR! thank you very much

PR #447, opened 2020-05-03
add k-d tree 

Vindaar:
This is almost a straight up port from scipy's KD tree (the Python implementation, not `cKDTree`, which is a C implementation). 

In principle the implementation works and gives the same results as the scipy tree.

Things left to do:
- [x] change return value from `seq[tuple]` to something more reasonable, maybe tuple of  `Tensor[T], Tensor[int]`. 
- [x] add specific query functions, like query ball
- [x] make use of real fancy indexing, now that it works
- [x] fix the `<` comparison of nodes
- [x] improve non-zero implementation
- [x] clean up code
- [x] add tests
- [x] add docs
- [ ] ~~maybe unify the metrics~~ used here with a general concept of metrics in arraymancer (given that e.g. dbscan allows for different metrics, maybe kmeans does too, haven't checked). NOTE: this will be done after the DBSCAN PR is merged (I'll try to get that done over the weekend). Then we can have a general `metrics` in `spatial` that can be used in clustering and here and by itself.


Vindaar:
Only have to turn a few scipy comparisons into a proper test now. 

Seems to work fine now. An example of a 2D tree based on 1000 random points as a non balanced tree:

kdtree

And the same input data to create a balanced tree (cutting at the median instead of the middle):

kdtree

Both are using the default leaf size of 16.

Vindaar:
Ok, aside from the question about if I should add the plotting / hyperrectangle stuff, this is done from my side. Since this is a bit bigger, I'd appreciate if someone could take a look. :)

Vindaar:
Ugh, I forgot about the annoying `-d:nimLegacyRandomInitRand` flag. Guess who's going to redo all the expected numbers now... :disappointed:

Vindaar:
Let's hope it works now.

Issue #445, opened 2020-04-29
fancy indexing: error with empty result


brentp:
```Nim
var r = @[1'f32, 1'f32].toTensor()
echo r[r <. 0]
```
gives: 
```
Error: unhandled exception: index out of bounds, the container is empty [IndexDefect]
```

Vindaar:
Hm, I tried to fix this right now, but I'm running into problems on how to handle empty tensors in arraymancer. 

So the underlying problem of this issue can be fixed easily by returning an empty tensor if the input has no true values / is an empty tensor / an empty seq. But then the resulting empty tensor has some weird behavior where it will run into all sorts of issues, because many of the low level operations expect the data storage to have elements in them. 

I'm not sure if the solution is to check in `p_accessors.nim` for size of the tensors too? I don't have a good overview on how those templates are used. Some input @mratsim would be appreciated.

Vindaar:
Just saw this issue: https://github.com/mratsim/Arraymancer/issues/210

mratsim:
Related: https://github.com/pytorch/pytorch/issues/5014

mratsim:
And seems like it's messy on Intel side as well: https://github.com/oneapi-src/oneDNN/issues/217

mratsim:
It seems like Numpy also IndexError on empty tensors: https://stackoverflow.com/questions/53747948/what-does-a-numpy-shape-starting-with-zero-mean

I'll try to copy numpy behavior on the basic reshaping and arithmetic operations implemented. Hopefully they make sense.

mratsim:
Now if someone can explain why this reliably returns 0.007812501848093234 ...

```
import numpy as np

def empty_add():
    empty = np.ndarray([])

    print(empty)

if __name__ == "__main__":
    empty_add()
```

mratsim:
Closed by #448

Issue #210, opened 2018-04-01
Support for completely empty tensors


mratsim:
Currently:

```Nim
import ../src/arraymancer

var a = newTensorUninit[float32](0)

echo a.repr
```
```
[shape = [data = [0, 0, 0, 0, 0, 0, 0],
len = 1],
strides = [data = [1, 0, 0, 0, 0, 0, 0],
len = 1],
offset = 0,
storage = [Fdata = 0x103b5d048[]]]
```

Displaying will crash
```
test_uninit.nim(14)      test_uninit
display.nim(23)          $
p_accessors.nim(138)     disp2d
data_structure.nim(211)  dataArray
system.nim(2833)         sysFatal
Error: unhandled exception: index out of bounds [IndexError]
```

PR #448, opened 2020-05-06
Empty tensors

mratsim:
This should handle the typical issues in #445 and #210

There doesn't seem to be a spec or at least an exhaustive document on all those corner behaviours so I went with gut feelings:
- For non-broadcasted operations, shape checks are unchanged, no special handling of empty tensors
- For broadcasted operations, only emptyTensor-by-scalar are handled and they return an empty tensor.
- Functions that take a single tensor as input and return a tensor will return an empty tensor (sum, min, max, variance on an axis).
  I didn't modify zeros_like and ones_like for now
- Function that take a single tensor as input and return a scalar are unchanged for now. Numpy behavior seems to return the neutral element, for example `np.sum` returns 0 and `np.prod` returns 1, I feel like this will lead to subtle bugs especially in iterative algorithm.
- Concatenation skips empty tensors
- selectors and fancy indexing returns empty tensors on empty input (tensor or mask or openarray of indices)
- Display does not crash anymore on empty tensors


src/tensor/aggregate.nim
```diff
@@ -18,19 +18,25 @@ import  ./backend/memory_optimization_hints,
...
+        ./private/p_empty_tensors,
         math
 import complex except Complex64, Complex32
 
 # ### Standard aggregate functions
 # TODO consider using stats from Nim standard lib: https://nim-lang.org/docs/stats.html#standardDeviation,RunningStat
 
+# Note: for aggregate that returns scalar, if the tensor is empty,
+#       Numpy seems to return the neutral element, do we want that?
+
```
Vindaar, line 30: From a purely mathematical point of view it does make sense. I see your point thought that it can introduce subtle bugs for cases where an empty tensor is actually the result of something that went wrong. By returning the identity that information is lost. 

But since the only real alternative I see is to raise an exception, I'm not sure. I'll think about it some more.

src/tensor/operators_logical.nim
```diff
@@ -17,6 +17,7 @@ import  ./data_structure,
         ./init_cpu
 
 # TODO tests
+# TODO Empty tensors boolean logic?
```
Vindaar, line 20: This is also difficult. I'd say that:
1. boolean operations only make sense if `a`, `b` have the same size. That would, in the current implementation using `map2_inline` automatically lead to an error if `a` was empty and `b` wasn't. That's imo fine.

Aside: currently `map2_inline` and friends only raise exceptions when compiled with `boundChecks` of course, so the errors might still be hard to reason about without checks, but that can't be helped?

2. if both tensors are empty, the result is also an empty tensor.

For a second I got confused by boolean tensor operations, which return `bool` and not `Tensor[bool]` where one could argue in terms of set theory for two empty tensors (-> two empty sets), but this doesn't apply here.

src/tensor/private/p_empty_tensors.nim
```diff
@@ -0,0 +1,45 @@
...
+  if isEmpty(t):
+    continue
+
+macro returnEmptyIfEmpty*(tensors: varargs[untyped]): untyped =
+  ## If any of the argument tensors are empty
+  ## return an empty tensor
+  result = newStmtList()
+  for tensor in tensors: # static for loop
+    let isEmptyCall = newCall(bindSym"isEmpty", tensor)
+    result = quote do:
```
Vindaar, line 43: This should be `result.add quote do:` no? Otherwise only the last tensor will be checked.

src/tensor/private/p_empty_tensors.nim
```diff
@@ -0,0 +1,45 @@
...
+      if `isEmptyCall`:
+        return
```
Vindaar, line 45: This relies on the fact that the default initialization of a tensor is empty. Is that wise? I see the problem that otherwise we need to know the type in order to create an empty tensor of the correct return type, I'm just wondering if the tensor will be initialized properly as an empty tensor otherwise.

src/tensor/selectors.nim
```diff
@@ -87,13 +90,17 @@ func masked_select*[T](t: Tensor[T], mask: Tensor[bool]): Tensor[T] {.noInit.} =
...
+  returnEmptyIfEmpty(t, mask)
   check_elementwise(t, mask)
 
   # TODO: fold_inline should accept an accumType like fold_axis_inline
   var size = 0
   for val in mask:
     size += int(val)
 
+  if size == 0:
+    return # Empty!
```
Vindaar, line 102: Same as my comment on `returnEmptyIfEmpty` return value.

src/tensor/selectors.nim
```diff
@@ -185,6 +194,9 @@ template masked_axis_select_impl[T](result: var Tensor[T], t: Tensor[T], mask: T
...
```
Vindaar, line 198: See above.

src/tensor/operators_logical.nim
mratsim: The checks should probably reenabled, at all times. That was at a time where I was too optimization-happy ;)

src/tensor/private/p_empty_tensors.nim
mratsim: Yes, I saw that before sleeping yesterday

src/tensor/private/p_empty_tensors.nim
mratsim: And empty tensor is defined as `size == 0`, all data accesses should lead to an error which is currently true with a seq backend.

I'll have to rework that when the backend is a pointer to add proper checks (same for OpenCL and Cuda).

One thing is the {.noInit.} proc that's true, I think I can solve that with some `typeof(result)()` hack

PR #464, opened 2020-07-10
Allow independent imports of all submodules

Vindaar:
Just played around trying to get independent imports to work. Turns out it's not that hard, if one reads the nimble readme with some attention, haha.

We have to add a `arraymancer` directory and need the submodule files in there, e.g. the main file for `tensor.nim`.

Now we can think about which parts of arraymancer can be used as submodules. I suppose in principle all that have their own submodules and are imported in `arraymancer.nim` via a single file. Not quite sure what that implies for the files in the `stats` dir. Need to test whether they can be imported separetely as:
```nim
import arraymancer/stats/kde
```
for instance.

~Sorry, that the gif diff looks so horrendeous.~

edit: turns out I accidentally staged the emacs backup files. Ooops, one sec.
edit2: much better

edit n: on my machine compiling the added `test_indep_import.nim` modified to use `import arraymancer` takes:
```sh
basti at void in ~/src/nim/arraymancer ツ nim c -f ./tests/test_indep_import.nim
...
Hint: 233051 lines; 1.882s; 226.945MiB peakmem; Debug build; proj: /home/basti/src/nim/arraymancer/tests/test_indep_import.nim; out: /home/basti/src/nim/arraymancer/tests/test_indep_import [SuccessX]
```
and with only importing tensor:
```sh
basti at void in ~/src/nim/arraymancer ツ nim c -f ./tests/test_indep_import.nim
...
Hint: 78384 lines; 0.801s; 118.441MiB peakmem; Debug build; proj: /home/basti/src/nim/arraymancer/tests/test_indep_import.nim; out: /home/basti/src/nim/arraymancer/tests/test_indep_import [SuccessX]
```

edit n+1: for ggplotnim the relative gains are more modest, but the absolute is about the same. On one recipe (the periodic table one, since it uses quite a few formulae), it goes from ~5.9 s down to ~ 4.7 s.


Vindaar:
Ok, now it should work with all submodules. The test also now includes a test of the (to me) unclear case of a submodule without a "main" submodule file, namely of the `stats` directory (it works!).

PR #475, opened 2020-11-25
fix warnings & deprecated functions

Clonkk:
This PR aims at fixing deprecated warnings when compiling.

* ``existsFIle`` -> ``fileExists``
* ``IndexError`` -> ``IndexDefect`` 
* ``csize`` -> ``csize_t``
* replace ``add`` for table by ``[]``

* Removed assert for 1D/2D Tensor on argmax

### DIFF
--- a/arraymancer.nimble
+++ b/arraymancer.nimble
@@ -1,11 +1,11 @@
 ### Package
-version       = "0.6.1"
+version       = "0.6.2"
 author        = "Mamy André-Ratsimbazafy"
 description   = "A n-dimensional tensor (ndarray) library"
 license       = "Apache License 2.0"
 
 ### Dependencies
-requires "nim >= 1.0.0",
+requires "nim >= 1.2.0",
   "nimblas >= 0.2.2",
   "nimlapack >= 0.1.1",
   "nimcuda >= 0.1.4",
--- a/arraymancer.nimble
+++ b/arraymancer.nimble
@@ -5,7 +5,7 @@ description   = "A n-dimensional tensor (ndarray) library"
 license       = "Apache License 2.0"
 
 ### Dependencies
-requires "nim >= 1.2.0",
+requires "nim >= 1.4.0",
   "nimblas >= 0.2.2",
   "nimlapack >= 0.1.1",
   "nimcuda >= 0.1.4",
--- a/src/arraymancer/tensor/aggregate.nim
+++ b/src/arraymancer/tensor/aggregate.nim
@@ -21,7 +21,7 @@ import  ./backend/memory_optimization_hints,
         ./algorithms,
         ./private/p_empty_tensors,
         math
-        
+
 import complex except Complex64, Complex32
 
 # ### Standard aggregate functions
@@ -174,7 +174,7 @@ proc argmax_max*[T](t: Tensor[T], axis: int): tuple[indices: Tensor[int], maxes:
   ##                                     [1],
   ##                                     [1]].toTensor
 
-  assert axis in {0, 1}, "Only 1D and 2D tensors are supported at the moment for argmax"
+  # assert axis in {0, 1}, "Only 1D and 2D tensors are supported at the moment for argmax"
   # TODO: Reimplement parallel Argmax (introduced by https://github.com/mratsim/Arraymancer/pull/171)
   #       must be done with care: https://github.com/mratsim/Arraymancer/issues/183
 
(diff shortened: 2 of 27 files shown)


Clonkk:
Due to IndexDefect usage, the CI needed to be updated to 1.4.0

src/arraymancer/tensor/aggregate.nim
```diff
@@ -174,7 +174,7 @@ proc argmax_max*[T](t: Tensor[T], axis: int): tuple[indices: Tensor[int], maxes:
   ##                                     [1],
   ##                                     [1]].toTensor
 
-  assert axis in {0, 1}, "Only 1D and 2D tensors are supported at the moment for argmax"
+  # assert axis in {0, 1}, "Only 1D and 2D tensors are supported at the moment for argmax"
```
Vindaar, line 177: This needs both a changelog entry and a test imo. Not sure if this should be part of this PR.

arraymancer.nimble
```diff
@@ -1,11 +1,11 @@
 ### Package
-version       = "0.6.1"
+version       = "0.6.2"
 author        = "Mamy André-Ratsimbazafy"
 description   = "A n-dimensional tensor (ndarray) library"
 license       = "Apache License 2.0"
 
 ### Dependencies
-requires "nim >= 1.0.0",
+requires "nim >= 1.4.0",
```
Vindaar, line 8: We could in principle define `IndexDefect` if we're running on an older Nim. I'm not sure if we need to throw out Nim 1.2 just because of Error vs Defect.

Vindaar (changes requested):
Thanks! Aside from my comments I agree with everything. Looks good to me.

I'd wait for @mratsim's input on my raised points as well.

src/arraymancer/tensor/aggregate.nim
Clonkk: Hmm, I wondered that as well but considering the issue associated to the assert in the comment has been closed I treated this assert as a deprecated warning that wasn't relevant anymore.

+1 for changelog & tests.

src/arraymancer/tensor/private/p_accessors_macros_desugar.nim
```diff
@@ -24,7 +24,7 @@ const Span = SteppedSlice(b: 1, step: 1, b_from_end: true)
 # #########################################################################
 # Slicing macros - desugaring AST
 
-macro desugar*(args: untyped): typed =
+macro desugar*(args: untyped): void =
```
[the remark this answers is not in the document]
Clonkk: This change is motivated by this warning : ``Warning: `typed` will change its meaning in future versions of Nim. `void` or no return type declaration at all has the same meaning as the current meaning of `typed` as return type declaration. [Deprecated]``

arraymancer.nimble
Clonkk: We absolutly can do that. I just try to avoid boilerplate code whenever I can, but if the 1.4.0 requirement is too harsh, then your suggesiton is better.

src/arraymancer/tensor/private/p_accessors_macros_desugar.nim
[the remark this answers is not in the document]
Vindaar: Oh, yes. I was expressing general confusion, because I only now noticed that this is what `desugar` returns.

arraymancer.nimble
Vindaar: Let's wait for mratsim's input here.

src/arraymancer/tensor/private/p_accessors_macros_desugar.nim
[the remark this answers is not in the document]
Vindaar: I've just looked at it again. What I would do is:
```suggestion
proc desugar*(args: NimNode): NimNode =
```
and then replace the calls of `desugar` which use `getAst` by regular calls. Works fine and is imo more clean.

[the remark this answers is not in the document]
Clonkk: I don't think this is relevant for this PR.

src/arraymancer/tensor/aggregate.nim
Clonkk: I'll add test and changelog this week I hope. I just had stuff to get done so couldn't take the time to take care of it.

Clonkk:
Since your PR rebasing the repo has been merged, I don't think this PR is relevant anymore. 
I can redo these modification (it does not take long) once a new version is released and we can take it from there. 
What do you think ?

mratsim:
It's still relevant, It's just that i wrapped the tests in a main() proc to avoid running out of globals as suite/test put everything in globals andwe only have 3000 or so globals allowed.

So all the changes in test files only need an extra indentation. But you can close and reopen a fresh one sure.

Clonkk:
I opened https://github.com/mratsim/Arraymancer/pull/480 as continuation of this one. I'd rather not deal with git conflict after a rebase.

PR #420, opened 2020-03-04
[Blocked by upstream bug] Change Tensor backend to pointer+length

mratsim:
This PR changes the CPU tensor backend to Laser pointer+length.

This would help on the issues tagged laser: https://github.com/mratsim/Arraymancer/issues?q=is%3Aissue+label%3ALaser+is%3Aopen

## Changing the backend

The current backend is using Nim sequences, this limits interoperability with other libraries and framework, in particular, having Arraymancer being just a view over a memory buffer described by pointer + length would allow zero-copy operations with NumPy, PyTorch and Tensorflow.

An additional benefit is enabling custom allocator (#419) which would be useful for distributed computing, NUMA-aware allocators and for situations where it's desirable to use the stack for example.

To provide deep immutability checks even with pointer object, a scheme using distinct type has been implemented:

https://github.com/mratsim/Arraymancer/blob/1a68830b14d3a8170d3918d0aba44369411ff205/src/laser/tensor/datatypes.nim#L12-L14

https://github.com/mratsim/Arraymancer/blob/1a68830b14d3a8170d3918d0aba44369411ff205/src/laser/tensor/datatypes.nim#L77-L89

https://github.com/mratsim/Arraymancer/blob/1a68830b14d3a8170d3918d0aba44369411ff205/src/laser/tensor/datatypes.nim#L105-L112

## Iterators

The current map_inline/map2_inline/map3_inline and apply_inline/apply2_inline/apply3_inline have several limits:
1. We need a variadic version that can handle arbitrary number of tensor to iterate onto. This limitation caused an inefficient iteration in GRU for example: https://github.com/mratsim/Arraymancer/blob/8acfd686b96a223dfe31803978b3915ab46e699f/src/nn_primitives/nnp_gru.nim#L76-L88

  Instead we could fuse all of those iterations and avoid intermediate tensors, similar to Intel Nervana Neon: https://github.com/NervanaSystems/neon/blob/8c3fb8a9/neon/layers/recurrent.py#L710-L723

2. The `x`, `y`, `z` arbitrary injected variables seem to come out-of-nowhere Laser forEach will lead to more readable code as the variables can use any name: https://github.com/numforge/laser/blob/d1e6ae6106564bfb350d4e566261df97dbb578b3/benchmarks/loop_iteration/iter_bench_prod.nim#L87-L90
  ```Nim
proc mainBench_libImpl(a, b, c: Tensor, nb_samples: int) =
  bench("Production implementation for tensor iteration"):
[...]

PR #477, opened 2020-11-27
Laser tensor PR rebase

Vindaar:
Here it is finally. Sorry for taking so long.

This is a rebase of the laser tensor PR #420 onto the current state of arraymancer. The major complication was the restructure of the whole arraymancer library to support importing individual parts as submodules.

I had to fix a few things that came up:
- there was a self assignment line in `pca.nim`, which caused the tensor to be destroyed maybe? I removed it and now it works with arc. Relevant commit 
: https://github.com/mratsim/Arraymancer/commit/63b470576a94afec4a264bdd9add7394e556feac
  Maybe my `=destroy` implementation is at fault? @clyybber maybe? 
- I had to add a `CpuStorageObj` for the `=destroy` proc signature. That's the same as would happen implicitly I assume?
- there's still the issue with the `supportsCopyMem` not supported in type sections and the workaround:
https://github.com/mratsim/Arraymancer/blob/laser-tensor-rebase/src/arraymancer/laser/tensor/datatypes.nim#L31-L36

As discussed on #420 this can cause issues where certain types (variant objects) match different branches in the type definition and in the code using `when T.supportsCopyMem`.
For that reason I will define a type class of types for which we know that they support `copyMem` which will then be used in the code as well as in the type definition to make sure laser based tensors are used for the vast majority of types used in practice.

I'll finish the latter maybe today and at the latest over the weekend. Wanted to open this PR first though, because I don't know how much more time I have today and to see what the CI says...


Vindaar:
CI failure is because I forgot to add `Complex` to `KnownSupportsCopyMem` from here:
https://github.com/mratsim/Arraymancer/blob/laser-tensor-rebase/tests/tensor/test_filling_data.nim#L33

However, it's a bit worrying that this failed in the first place, indicating that non mem copyable types may have hidden bugs. It could be a good idea to run all basic tests in a mode where all types are treated as mem copyable? Maybe have a 
```nim
when not defined(noPtrLenBackend):
type
  KnownSupportsCopyMem* = SomeNumber | char | Complex[float64] | Complex[float32]
else:
  KnownSupportsCopyMem* = distinct object
```
and then run tests with `-d:noPtrLenBackend` in addition?

Vindaar:
FYI: https://github.com/nim-lang/Nim/issues/16185 is still an open issue, which makes the code here unsafe if used with ARC. 

And a personal blocker: `ggplotnim` does not pass its dataframe tests. It throws an `out of memory` error. I'll try to debug that in the coming days. Once I figure out if it's a problem on ggplotnim's or arraymancer's side, I'd consider merging this PR. @mratsim still has to review it at that point though.

mratsim (approved):

PR #479, opened 2020-12-12
[WIP] Laser forEach iterators

mratsim:
This replaces the use of `mapX_inline` and `applyX_inline` by the forEach / forEachContiguous / forEachParallel / forEachSerial laser iterators.

This is particularly valuable for recurrent neural network like GRU because we can implement the equation in a straightforward manner with meaning ful variable name instead of magic `x`, `y`, `z` and we would have needed an `apply11_inline` anyway (with the correct var/non-var parameters).

TODO:
-  There is a significant parallel performance regression on GRU when running test_nnp_gru. While before this PR the parallel version was "only" 2x slower than serial, now it's 13x slower than serial which probably signals a false sharing issue.
  Even then RNN are a type of iterative stencil application that require special care and often used for polyhedral benchmarking so another approach using tiling is probably needed to properly speed those up. (see https://github.com/numforge/laser/blob/master/research/automatic_loop_nest_scheduling.md#polyhedral-approaches)

Not in scope:
- using forEach for backward propagation in GRU: this is a headache inducing refactoring

### DIFF
--- a/benchmarks/ex01_xor.nim
+++ b/benchmarks/ex01_xor.nim
@@ -1,50 +1,52 @@
 import ../src/arraymancer
 
 # Learning XOR function with a neural network.
+proc main() =
+  # Autograd context / neuralnet graph
+  let ctx = newContext Tensor[float32]
+  let bsz = 32 # batch size
 
-# Autograd context / neuralnet graph
-let ctx = newContext Tensor[float32]
-let bsz = 32 # batch size
+  let x_train_bool = randomTensor([bsz * 100, 2], 1).astype(bool)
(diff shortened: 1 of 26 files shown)


mratsim:
Some changes in operator BLAS L1, require changing autograd to `+.=` because the `+=` based on apply2_inline was doing implicit broadcast somehow. But then we have cascading issues that requires changing `+.=` to use broadcast2 and broadcast2 fixes, and then we have the `Stack` autograd layer that is failing tests. See https://github.com/mratsim/Arraymancer/commit/2929cb5a62ec00ad951c10bb8d1a0db8bd08fe09 in the laser-iterators-stashed branch

PR #480, opened 2020-12-13
Fix some deprecation

Clonkk:
* csize ->csize_t
* IndexError -> IndexDefect : Do we need to introduce an IndexDefect ``when NimVersion <= 1.4.0`` or do webump the CI version to 1.4 ?
* Table add -> []=
* ``.==`` -> ``==.``
* macro ``untyped`` ->  ``void``
* Removed assert 1d/2d for argmax_max; kept reference on issues / PR in changelog


Vindaar (approved):
Looks great to me!

@mratsim Did you wish to test Nim version 1.0.6 to make sure we keep supporting it?

mratsim:
I think it's important because the newer version broke nvcc /Cuda.

Clonkk:
I'll update the PR tonight to reflect this

Clonkk:
I added a file ``std_version_types`` that is destined to contains types introduced by newer version of Nim. 
That way it is easy to manage  even if more deprecation come with Nim 1.4.4, 1.6.0, etc.

Vindaar (approved):
Still looks good to me!

I think `std_version_types` is a good solution. Even if no further deprecations / additions were to happen at least this encapsulates those things in one place!

mratsim (approved):

Issue #459, opened 2020-05-26
Ambiguous call when using tanh activation layer


dylanagreen:
The following code block is a minimal example that reproduces this issue:
```nim
import arraymancer

let (D_in, D_out) = (5, 1)
var ctx = newContext Tensor[float32]

network ctx, TestNet:
  layers:
    fc1: Linear(D_in, D_out)
  forward x:
    x.fc1.tanh

let model = ctx.init(TestNet)
```

Arraymancer compiled at #head (974efe7a57481c06d4fc6995629e99603e33c017 as of writing) using nimble produces the following error when trying to use a tanh activation layer:

```
template/generic instantiation of `network` from here 
template/generic instantiation of `tanh` from here
Error: ambiguous call; both ufunc.tanh(t`gensym13805012: Tensor[tanh.T]) [declared in /Users/dylan/.nimble/pkgs/arraymancer-#head/tensor/ufunc.nim(94, 15)] and nnp_activation.tanh(t: Tensor[tanh.T]) [declared in /Users/dylan/.nimble/pkgs/arraymancer-#head/nn_primitives/nnp_activation.nim(38, 6)] match for: (Tensor[system.float32])
```

Here is the output of `nim --version`

```
Nim Compiler Version 1.2.0 [MacOSX: amd64]
Compiled at 2020-04-22
Copyright (c) 2006-2020 by Andreas Rumpf
```

This error does not appear when arraymancer is installed at 0.6.0 (which is mislabeled as 0.5.2 in nimble), and only appears if you compile or install arraymancer at the most recent commit.

dylanagreen:
I can confirm that this still happens when compiling arraymancer at HEAD (now bdcdfe1).

Vindaar:
Feel free to reopen if the issue still persists after all.

PR #484, opened 2020-12-29
Fix issue #459

Vindaar:
Fixes issue #459 by choosing an explicit overload.

I'm not sure why this issue did not happen at some point in the past.

@mratsim I was a bit confused about choosing `ufunc`: The `value` in a `Variable` is *always* a `Tensor`, right?


Vindaar:
Oookayyy. So apparently the failed CI is the reason for the regression. Why is `ufunc` not visible on Nim 1.0.6? Will take a look...

Vindaar:
Rebased and attempting to fix for old nim.

Issue #491, opened 2021-01-04
Export `cumsum` procedure and expand to work on n-dim Tensors


filipeclduarte:
I have three suggestions: 

1. Export `cumsum` procedure: 
 https://github.com/mratsim/Arraymancer/blob/1a2422a1e150a9794bfaa28c62ed73e3c7c41e47/src/arraymancer/ml/clustering/kmeans.nim#L26
```nim
proc cumsum[T: SomeFloat](p: Tensor[T]): Tensor[T] {.noInit.} =
  ## Calculates the cumulative sum of a vector.
  ## Inputs:
  ##  - p: a rank-1 tensor to cumulatively sum
  ##
  ## TODO: implement parallel prefix sum
  ## See: https://en.wikipedia.org/wiki/Prefix_sum#Algorithm_2:_Work-efficient
  ##
  ## Returns:
  ##  - A tensor cumulatively summed, that is, add each value to
  ##    all previous values, sequentially
  result = p.clone()
  assert p.rank == 1
  assert p.shape[1] == 0
  let n_rows = p.shape[0]
  for i in 1..<n_rows:
    result[i] += result[i-1]
```

2. Generalize this procedure to work on n-dimensional Tensors.

3. Create a procedure to do cumulative product `cumprod` that is similar to `cumsum`, but instead of sum, use prod.

filipeclduarte:
@HugoGranstrom suggested this proc `cumprod` at the discord: 

```nim
proc cumsum*[T](t: Tensor[T], axis: int): Tensor[T] =
  ## Calculates the cumulative sum of a rank-n Tensor.
  ## Inputs:
  ##  - t: a rank-n tensor to cumulatively sum
  ##  - axis: int
  ## Returns:
  ##  - A tensor cumulatively summed at axis
  result = zeros_like(t)
  for i, tAxis in enumerateAxis(t, axis):
    var temp = result.atAxisIndex(axis, i)
    if i == 0:
      temp[_] = tAxis
    else:
      temp[_] = result.atAxisIndex(axis, i-1) + tAxis
``` 
Then I suggested to create the `cumprod` as:

```nim 
proc cumprod*[T](t: Tensor[T], axis:int): Tensor[T] = # from hugogranstrom
  ## Calculates the cumulative product of a rank-n Tensor.
  ## Inputs:
  ##  - t: a rank-n tensor to cumulatively sum
  ##  - axis: int
  ## Returns:
[...]

PR #492, opened 2021-01-05
Cumsum cumprod

filipeclduarte:
Following the issue: https://github.com/mratsim/Arraymancer/issues/491 @HugoGranstrom helped me develop those two procedures: `cumsum` and `cumprod`, which are very useful to scientific computing and finance.

### DIFF
--- a/src/arraymancer/tensor/aggregate.nim
+++ b/src/arraymancer/tensor/aggregate.nim
@@ -15,6 +15,7 @@
 import  ./data_structure,
         ./init_cpu,
         ./higher_order_foldreduce,
+        ./operators_broadcasted,
         ./math_functions,
         ./accessors,
         ./algorithms,
--- a/src/arraymancer/tensor/aggregate.nim
+++ b/src/arraymancer/tensor/aggregate.nim
@@ -245,3 +245,39 @@ proc iqr*[T](t: Tensor[T]): float =
   let tS = t.sorted
   result = percentile(tS, 75, isSorted = true) -
            percentile(tS, 25, isSorted = true)
+
+proc cumsum*[T](t: Tensor[T], axis:int): Tensor[T] = # from hugogranstrom
+  ## Calculates the cumulative sum of a rank-n Tensor.
+  ## Inputs:
+  ##  - t: a rank-n tensor to cumulatively sum
+  ##  - axis: int
+  ## Returns:
+  ##  - A tensor cumulatively summed at axis, that is, add each value to
+  result = zeros_like(t)
+  for i, tAxis in enumerateAxis(t, axis):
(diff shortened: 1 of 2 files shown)


mratsim:
Thanks! Can you add a test in the test suite?

Ultimately we likely want that parallelized and a test would be helpful to ensure that we don't mess it up.

filipeclduarte:
I added the test for these procedures. I'm gonna commit the changes.

src/arraymancer/tensor/aggregate.nim
```diff
@@ -245,3 +245,33 @@ proc iqr*[T](t: Tensor[T]): float =
...
   result = percentile(tS, 75, isSorted = true) -
            percentile(tS, 25, isSorted = true)
+
+proc cumsum*[T](t: Tensor[T], axis:int): Tensor[T] = # from hugogranstrom
+  ## Calculates the cumulative sum of a rank-n Tensor.
+  ## Inputs:
+  ##  - t: a rank-n tensor to cumulatively sum
+  ##  - axis: int
+  ## Returns:
+  ##  - A tensor cumulatively summed at axis, that is, add each value to
```
mratsim, line 255: can you add `mixin _` here, it seems like the `_` symbol is not visible in some cases and breaks CI

src/arraymancer/tensor/aggregate.nim
```diff
@@ -245,3 +245,33 @@ proc iqr*[T](t: Tensor[T]): float =
...
+    else:
+      temp[_] = result.atAxisIndex(axis, i-1) + tAxis
+
+proc cumprod*[T](t: Tensor[T], axis:int): Tensor[T] = # from hugogranstrom
+  ## Calculates the cumulative sum of a rank-n Tensor.
+  ## Inputs:
+  ##  - t: a rank-n tensor to cumulatively sum
+  ##  - axis: int
+  ## Returns:
+  ##  - A tensor cumulatively summed at axis, that is, add each value to
```
mratsim, line 270: and similar here can you add `mixin _` here, it seems like the `_` symbol is not visible in some cases and breaks CI

mratsim (approved):
Nim's early symbol resolution strikes again :/ https://github.com/nim-lang/Nim/issues/8677 https://github.com/nim-lang/Nim/issues/6387

src/arraymancer/tensor/aggregate.nim
filipeclduarte: added

src/arraymancer/tensor/aggregate.nim
filipeclduarte: added

PR #499, opened 2021-03-09
Add unsafe_raw_data

Clonkk:
This PR add a simple unsafe_raw_data access function to get a ``ptr UncheckedArray[T]`` based on unsafe_raw_offset.

No distinction is made so far on Immutable vs Mutable (since this is not intended to write to this ptr anyway, there is ``fromBuffer`` for that). 
Other possibility would be:  
* forbid this func for non-var Tensor 
* define 2 func :

```nim
func unsafe_raw_data*[T: KnownSupportsCopyMem](t: Tensor[T], aligned: static bool = true): UncheckedArray[T] {.inline.} 

func unsafe_raw_mut_data*[T: KnownSupportsCopyMem](t: var Tensor[T], aligned: static bool = true): UncheckedArray[T] {.inline.} 
```

Since ``unsafe_raw_data`` directly use ``unsafe_raw_offset().distinctBase()`` I didn't see the need to add another unit test; but if It's a wrong assumption, I'll add one .

I added the 0.6.2 release date in the changelog, but if you want in another PR i can just revert the commit and open a new one

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,7 @@
 Arraymancer v0.6.X
 =====================================================
 Changes :
-  - Add ``unsafe_raw_data`` as replacement of ``dataArray`` to return a ``ptr UncheckedArray`` 
+  - Add ``toUnsafeView`` as replacement of ``dataArray`` to return a ``ptr UncheckedArray``
 
 
 Arraymancer v0.6.2 Dec. 22 2020
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,13 @@
-Arraymancer v0.x.x
+Arraymancer v0.6.X
+=====================================================
+Changes :
+  - Add ``unsafe_raw_data`` as replacement of ``dataArray`` to return a ``ptr UncheckedArray`` 
+
+
+Arraymancer v0.6.2 Dec. 22 2020
 =====================================================
 
-Changes (TODO):
+Changes :
   - Fancy Indexing (#434)
   - ``argmax_max`` obsolete assert to 1D/2D removed. It refered to issues #183 and merged PR #171
 
--- a/src/arraymancer/laser/tensor/datatypes.nim
+++ b/src/arraymancer/laser/tensor/datatypes.nim
@@ -177,6 +177,17 @@ func unsafe_raw_offset*[T: KnownSupportsCopyMem](t: var Tensor[T], aligned: stat
   ## and that the data is aligned by LASER_MEM_ALIGN (default 64).
   unsafe_raw_offset_impl(t.offset)
 
+func unsafe_raw_data*[T: KnownSupportsCopyMem](t: Tensor[T], aligned: static bool = true): UncheckedArray[T] {.inline.} =
+  ## Returns a ``ptr UncheckedArray`` to the start of the valid data.
+  ## Avoid writing to the returned pointer. Use ``fromBuffer`` instead to write Tensor data from a pointer.
+  ##
+  ##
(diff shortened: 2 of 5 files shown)


src/arraymancer/laser/tensor/datatypes.nim
```diff
@@ -177,6 +177,17 @@ func unsafe_raw_offset*[T: KnownSupportsCopyMem](t: var Tensor[T], aligned: stat
   ## and that the data is aligned by LASER_MEM_ALIGN (default 64).
   unsafe_raw_offset_impl(t.offset)
 
+func unsafe_raw_data*[T: KnownSupportsCopyMem](t: Tensor[T], aligned: static bool = true): UncheckedArray[T] {.inline.} =
```
Vindaar, line 180: Will comment more later, but this should be `ptr UncheckedArray[T]` as the return value, surely?

Clonkk: I can't believe I've overlooked that

Vindaar:
On the one hand I feel that having an overload that removes the (imo pretty nice) Raw(Im)MutableView type and its theoretical safety is a net loss. But on the other hand:
1. `distinctBase` is one of those not really well known procs, so less experienced people trying to use arraymancer will certainly stumble over it
2. `Raw(Im)MutableView` is not as easy to work with (from the top of my head I'm not even sure if we have accessors for it borrowed? Well, probably yes)
3. For interop with other data sources these types won't be recognized anyway and hence all ideas of type safety are kind of out of the window anyway.

So yeah, I think it's a useful addition. For arraymancer internals we should keep using the other variants though, imo. 

Not sure if another name might be more useful for this to cause confusion between the other `unsafe_raw_*` funcs and this though.

Clonkk:
Ping @Vindaar @mratsim  :)

Vindaar:
I think `toUnsafeView` is a good name. Fits the typical Nim naming scheme (e.g. `toOpenArray`), contains `unsafe` and `view` highlighting that it's an unsafe no copy operation. :)

Clonkk:
Do you think this can be merged ?

src/arraymancer/tensor/data_structure.nim
```diff
@@ -185,7 +185,7 @@ proc get_offset_ptr*[T](t: CudaTensor[T] or ClTensor[T]): ptr T {.noSideEffect,
   ##     - A pointer to the offset start of its data
   t.storage.Fdata[t.offset].unsafeAddr
 
-proc dataArray*[T: KnownSupportsCopyMem](t: Tensor[T]): ptr UncheckedArray[T] {.noSideEffect, inline, deprecated: "Use unsafe_raw_offset instead".}=
+proc dataArray*[T: KnownSupportsCopyMem](t: Tensor[T]): ptr UncheckedArray[T] {.noSideEffect, inline, deprecated: "Use unsafe_raw_data instead".}=
```
mratsim, line 188: `toUnsafeView`

changelog.md
```diff
@@ -1,7 +1,13 @@
-Arraymancer v0.x.x
+Arraymancer v0.6.X
 =====================================================
+Changes :
+  - Add ``unsafe_raw_data`` as replacement of ``dataArray`` to return a ``ptr UncheckedArray``
```
mratsim, line 4: toUnsafeView

mratsim (commented):
looks good to me, though comments/changelog need to reflect the new name

PR #503, opened 2021-04-02
Implemented rudimentary Graph Convolutional Network

anon767:
Hi,
I implemented a rudimentary GCN as described here: https://arxiv.org/abs/1609.02907, as it may be of interest for some.

The GCN gate calculates (AX)W, where A is the adjacency matrix of the input graph and the result of the matrix multiplication of A and node features X are simply passed into a linear layer.
An End-to-End test is included and the DSL is adjusted accordingly.


mratsim:
Sorry for disappearing, that looks awesome.

Issue #500, opened 2021-03-12
Two (small) issues with display of tensor


pietroppeter:
example ([playground](https://play.nim-lang.org/#ix=2SAK)):
```nim
import arraymancer, sequtils
echo toSeq(1..24).toTensor().reshape(2,3,4)
```
output:
```
Tensor[system.int] of shape [2, 3, 4]" on backend "Cpu"
| | 	1	2	3	4 | 	13	14	15	16|
| | 	5	6	7	8 | 	17	18	19	20|
| | 	9	10	11	12 | 	21	22	23	24|
```

The two issues are:

1. an additional `"` after tensor shape
2. and additional set of `|` in first columns

First one is a trivial fix ([source](https://github.com/mratsim/Arraymancer/blob/master/src/arraymancer/tensor/display.nim#L21)); the second one might be a little be more involved ([source](https://github.com/mratsim/Arraymancer/blob/9a575de9e56fe3f5c10a5315765f5ded0332c44b/src/arraymancer/tensor/private/p_display.nim#L84)), but I could try to produce a PR with a fix for both.

Also, not sure what should be the expected display of a 4d tensor, an [example](https://play.nim-lang.org/#ix=2SAT):
```nim
import arraymancer, sequtils
echo toSeq(1..16).toTensor().reshape(2,2,2,2)
```
output:
```
Tensor[system.int] of shape [2, 2, 2, 2]" on backend "Cpu"
|						|
|-	-	-	-	-	-	-|
| | 	1	2 | 	5	6|
| | 	3	4 | 	7	8|
|-	-	-	-	-	-	-|
| | 	9	10 | 	13	14|
| | 	11	12 | 	15	16|
```

I noticed these two issues while writing a [remake of the Tutorial](https://pietroppeter.github.io/nblog/drafts/arraymancer_tutorial.html) using a library I am developing: [nimib](https://github.com/pietroppeter/nimib).

I also noticed this very interesting issue #488 which is relevant to documentation and for which maybe nimib in the future (I am close to first release of 0.1) might be able to help with.

mratsim:
Unfortunately 4D display on a 2D screen is a bit hard. Ideally 4D display looks like that https://github.com/mratsim/Arraymancer/issues/5#issuecomment-751852128

![image](https://user-images.githubusercontent.com/22738317/111026690-de982c00-83eb-11eb-9c2a-e3a83df429f3.png)

I try to reverse engineer the Haskell code of that library but got lost in Monads :/.

[...]

PR #509, opened 2021-06-30
ND tensor printing

Vindaar:
(will write in a few min)

edit:

Well, a few minutes quickly turn into an hour, haha.

This finally improves the pretty printing of an ND tensors. Instead of special casing each dimension, this is a generic approach that should scale to N dimensions. I've only tested to 5D (see tests), but that works fine. At even higher dimensions the algorithm should work correctly, but it's possible that the layouting of separator lines might not look perfectly.

The idea is simple:
- start with an ND tensor
- if N > 2 take axis 0 and iterate over that axis. Recurse on the tensor slice
  - if current slice is of odd rank, stack the result horizontally (next to each other)
  - if current slice is of even rank, stack the result vertically (on top of each other)
- once N = 2, use the existing (slightly modified) 2D pretty printing code 

In addition we add separators between the vertical / horizontal splits based on the current rank. On the outside of each printed block, the current index of that axis is printed.
The required alignment of the fields is performed by an initial string conversion of all elements and picking the largest string as a reference. If one stores a novel in a `Tensor[string]` it'll thus be ugly. :P This leads to perfect alignment independent of the scale of numbers stored. If very large and very small numbers appear it leads to a lot of whitespace between the smalll numbers. I preferred that to aligning only sub tensors.

There remains one question: the Haskell implementation referenced by @mratsim in #5 and #500 prints the number of the axis instead of the index along that axis on the outside. I personally prefer the approach I implemented as it allows to read off the exact index even in large outputs rather easily without manual counting. It does require one to think about which number corresponds to what axis though.

A note on performance: pretty printing a tensor is *slow*. I did not make (nor did the existing code) make any attempts at making sure all the string operations are particularly fast!

An example:

```
let t1 = toSeq(1..144).toTensor().reshape(2,3,4,3,2)
echo t1
# Tensor[system.int] of shape "[2, 3, 4, 3, 2]" on backend "Cpu"
#                           0                            |                            1
#        0            1            2            3        |         0            1            2            3
#   |1        2| |7        8| |13      14| |19      20|  |    |73      74| |79      80| |85      86| |91      92|
# 0 |3        4| |9       10| |15      16| |21      22|  |  0 |75      76| |81      82| |87      88| |93      94|
#   |5        6| |11      12| |17      18| |23      24|  |    |77      78| |83      84| |89      90| |95      96|
#   ---------------------------------------------------  |    ---------------------------------------------------
#        0            1            2            3        |         0            1            2            3
#   |25      26| |31      32| |37      38| |43      44|  |    |97      98| |103    104| |109    110| |115    116|
# 1 |27      28| |33      34| |39      40| |45      46|  |  1 |99     100| |105    106| |111    112| |117    118|
#   |29      30| |35      36| |41      42| |47      48|  |    |101    102| |107    108| |113    114| |119    120|
#   ---------------------------------------------------  |    ---------------------------------------------------
#        0            1            2            3        |         0            1            2            3
#   |49      50| |55      56| |61      62| |67      68|  |    |121    122| |127    128| |133    134| |139    140|
# 2 |51      52| |57      58| |63      64| |69      70|  |  2 |123    124| |129    130| |135    136| |141    142|
#   |53      54| |59      60| |65      66| |71      72|  |    |125    126| |131    132| |137    138| |143    144|
#   ---------------------------------------------------  |    ---------------------------------------------------
```

edit2:

As the CI is still failing, maybe it's a good idea to make another PR to fix it first and then rebase.
Aside from that I should probably add a pretty printing `Tensor[float]` example as well.

edit3:

Ah, another thing I forgot to mention: If we consider merging this, should I remove the existing `disp3d` and `disp4d` procedures?


Vindaar:
Ok, I've cleaned it up a bit now. Removed the `disp3d` and `disp4d` proc, added a custom precision argument for floating point tensors and restructured the `$` to use `pretty`. 

Also applied the changes to CUDA tensors. From my side this seems done for now. 

One idea I have that could improve it further: Allow coloring the separators `|` and `-` for better visibility. But I'm not sure if it's a good idea to apply color to these always.

Vindaar:
And now I've added a tutorial section explaining the pretty printing. now I'm actually done.

Vindaar:
Ah, and of course we immediately run into regressions of float printing between Nim version 1.0 and current devel. I'll put the float printing tests behind a nim version check.

Vindaar:
Rebased onto current master to resolve merge conflict about slicing tutorial.

mratsim:
Thanks, that' awesome!

PR #515, opened 2021-07-05
Fix examples & compile in CI to avoid regressions

### LINKED ISSUE #514: shakespeare_generator doesn't compile on Nim 1.4.8 / Arraymancer 0.7.1
Compilation of [this example](https://github.com/mratsim/Arraymancer/blob/master/examples/ex06_shakespeare_generator.nim) fails with an [error](https://github.com/mratsim/Arraymancer/blob/master/examples/ex06_shakespeare_generator.nim#L59): 

`(59, 6) Error: 'strToTensor' can have side effects`

This appears to be caused by a `newTensor` call adding side effects. Making all the `func`s `proc`s in the example fails with a different error: `'optimizerAdam' can have side effects`.

[This line](https://github.com/mratsim/Arraymancer/blob/master/examples/ex06_shakespeare_generator.nim#L295) also gives me a "wrong number of arguments" / "illformed AST" error in VS Code, but I'm not sure how accurate it is with other errors blocking compilation.

This is my first time trying out the library, so I'm not sure if the examples are outdated or whether I'm doing something wrong. Are they tested in CI? If not, it might be worth ensuring that they still compile with each new release/breaking change.

```
Nim Compiler Version 1.4.8 [Linux: amd64]
arraymancer  [0.7.1]
Command: nim r -d:danger <example_file>
```

Vindaar:
This fixes all examples to make them compile on current devel. Let's hope this doesn't break the CI for old Nim versions (but who are we kidding). 

Fixes #514 if it all works.

### DIFF
--- a/arraymancer.nimble
+++ b/arraymancer.nimble
@@ -143,6 +143,15 @@ proc test(name, switches = "", split = false, lang = "c") =
     exec "nim " & lang & " -o:build/" & name & switches & " -r tests/" & name & ".nim"
   else:
     exec "nim " & lang & " -o:build/" & name & switches & " -r tests/_split_tests/" & name & ".nim"
+  # try to compile (not run) all examples to avoid regressions
+  let examples = @["ex01_xor_perceptron_from_scratch.nim",
+                   "ex02_handwritten_digits_recognition.nim",
+                   "ex03_simple_two_layers.nim",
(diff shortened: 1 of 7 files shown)


Vindaar:
Wow, it actually works fine. :partying_face:

PR #518, opened 2021-07-09
DBSCAN rebase using k-d tree

Vindaar:
This is a rebase of the PR by @abieler #413. The credit goes to him, even if there's only 2 commits of his visible here.

Currently this is a draft, as the code is based on the k-d tree PR #447 to make use of the performance improvements achievable for nearest neighbor computation. Also it allows to unify the metrics into a single `distances.nim` file.

Things left to do:
- [x] change metrics to use an `typedesc[T]` calling convention
- [x] make k-d tree use the metrics defined in `distances.nim`
- [x] replace nearest neighbor computation based on distance matrix by k-d tree
- [x] write DBSCAN tests
- [x] finish DBSCAN documentation

An example of a clustering on a set of randomly generated clusters using DBSCAN:

dbscan_example

I'll finish the remaining things tomorrow. :)


Vindaar:
I consider this done for the time being as well. 

One might want to add more test cases for DBSCAN in the future.

abieler:
:tada: glad i could contribute a bit. did you do a benchmark vs the sklearn method in terms of execution speed?

Vindaar:
>  glad i could contribute a bit. did you do a benchmark vs the sklearn method in terms of execution speed?

No, I didn't do any benchmarking yet. Making use of the k-d tree sped things up significantly, but it may very well still be comparatively slow due to the way the query ball procedure for the k-d tree works at the moment. 

I personally just wanted to have an implementation that is working to test on my data for my PhD.

Feel free to play around of course!

Issue #523, opened 2021-08-11
Cannot clone Tensors of enum type


XDEmer0r-L0rd-360-G0d-SlayerXD:
As the name implies, there are issues with cloning some tensors.
My use case: 
```nim
import arraymancer

type
    Nest = enum
        a, b, c
    Change = Tensor[Nest]

proc layer(t: Change): Tensor[Nest] =
    return t.clone()

var ar = newTensor[Nest]([3, 3])
echo ar.layer()
```
And my simplified case:
```nim
import arraymancer

type
    Nest = enum
        a, b, c

var ar = newTensor[Nest]([3, 3])
echo ar.clone()
```

Both give an error of
```
d:\dump\Git\Hydris\temp.nim(10) temp
C:\Users\Nicholas\.nimble\pkgs\arraymancer-0.7.5\arraymancer\tensor\init_copy_cpu.nim(27) clone
C:\Users\Nicholas\.nimble\pkgs\arraymancer-0.7.5\arraymancer\tensor\higher_order_applymap.nim(85) contiguousImpl
C:\Users\Nicholas\.nimble\pkgs\arraymancer-0.7.5\arraymancer\tensor\private\p_accessors.nim(100) atIndexMut
C:\Users\Nicholas\.nimble\pkgs\arraymancer-0.7.5\arraymancer\tensor\private\p_accessors.nim(63) getIndex
C:\Users\Nicholas\.nimble\pkgs\arraymancer-0.7.5\arraymancer\tensor\private\p_checks.nim(32) check_index
Error: unhandled exception: Number of arguments: 1, is different from tensor rank: 2 [IndexDefect]
Error: execution of an external program failed: 'd:\dump\Git\Hydris\temp.exe '
```
despite a sucessfull compilation.

Am I doing something wrong, or is this a bug? I am using nim 1.4.8 with arraymancer 0.7.5

-Cheers!

Vindaar:
Oh, this is very interesting. I'll have a look later today.

Vindaar:
This should be fixed in `v0.7.6`, which I just tagged. Feel free to reopen, if something is still not working!

PR #524, opened 2021-08-13
Fix issue #523 by fixing `map_inline` for non memcopyable rank > 1

Vindaar:
The underlying issue of the failed `clone` in issue #523 is due to `map_inline` being ill defined for tensors of rank > 1 that store types that are not `KnownSupportsCopyMem` (our custom type class of types we *know* we can `copyMem`).

`enumerate` yields the actual indices (it is an enumeration after all) of the contiguous memory location. By accessing `[i]` for a rank > 1 tensor, it is an invalid access to the tensor. Instead (for now) we just access the underlying `seq[T]` store directly at the corresponding index.

In this case some way to set the memory location directly would be very convenient (that is independent of the rank & mem-copyability). Something like `atIndexMut`, but taking a single index. This goes back to the "making `atIndex` and friends public" etc.


Vindaar:
We seriously have to take out that MNIST test. It breaks like in 50% of the cases.

PR #525, opened 2021-09-04
added example for saving/loading models

Niminem:
<3


examples/ex_07_save_load_model.nim
```diff
@@ -0,0 +1,93 @@
+import arraymancer, strformat, os
+
+discard """
```
Vindaar, line 3: Instead of using `discard """` you can just either use double `##` or straight `#[` and `]#` to open and close a multi line comment.
This has the advantage that it's a proper documentation comment and would show up if running the file through `nim doc` (even though that's not important for these examples).

Vindaar (changes requested):
Just a minor thing here. Aside from that could you add this example to the test suite by adding it here:

https://github.com/mratsim/Arraymancer/blob/master/arraymancer.nimble#L157-L162

?

Otherwise thanks a lot!

examples/ex_07_save_load_model.nim
Niminem: Thanks @Vindaar I've replaced it with a simple multiline comment

Niminem: @Vindaar I've also added the example to the .nimble file

Niminem:
> Just a minor thing here. Aside from that could you add this example to the test suite by adding it here:
> 
> https://github.com/mratsim/Arraymancer/blob/master/arraymancer.nimble#L157-L162
> 
> ?
> 
> Otherwise thanks a lot!

@Vindaar  Added to test suite

examples/ex07_save_load_model.nim
```diff
@@ -0,0 +1,93 @@
+import arraymancer, strformat, os
```
Vindaar, line 1: You need to import it using relative imports, see the other examples.

Vindaar:
Not sure what the issue is on 1.4. Need to check that locally.

examples/ex07_save_load_model.nim
Niminem: resolved

Vindaar:
I don't quite understand what changed so that the 1.4 CI doesn't pass anymore (on any of the NN examples). 

Does the code still compile for you locally? For me on devel from a few days ago it doesn't. Maybe some fix was backported to 1.4 so that now it fails?

Niminem:
Glad you found it, I had no idea lol. I ran only ran it on 1.4.8 and devel before I made the PR

Niminem (commented):
fetched upstream, updated spelling error w/ model loading proc, tested on nim versions 1.4 and above

Niminem:
@Vindaar she should be ready to pull now

Vindaar:
Nice, thank you! 

One thing that could be nice is to add a couple of lines as a sanity test. I.e. after the save call reset the network, reload from the files and then re-predict some specific input and compare with that same sample predicted before. Just as a way to show that indeed, saving & loading works as expected.

But I'll merge it like this, if you want.

Niminem:
You're welcome @Vindaar! I'll add a quick sanity test today, I think that's a great idea

Niminem:
@Vindaar I've added a quick sanity check that simply compares the weights of the hidden layer (model 1) to the weights of the hidden layer (model 2)

`var hidden_weights = model.hidden.weight.value
let newModel = ctx.load()
doAssert newModel.hidden.weight.value == hidden_weights, "loaded model weights do not match with original model"`

Vindaar:
Thanks a lot for this again!

PR #528, opened 2021-10-14
fix autograd code by keeping proc signatures unspecified

Vindaar:
By keeping the procedure signatures exactly as the `Backward[TT]` proc
signature requires and then converting the input argument to the
correct type, we work around the recent compiler change, as mentioned
in:

https://github.com/nim-lang/Nim/issues/18828#issuecomment-916021277

Sorry @Niminem it took so long. I forgot about this issue.


Niminem:
@Vindaar no worries!

Issue #543, opened 2021-12-29
einsum in proc fails when result is a Tensor


asnt:
Follow-up of #537.
#537 fixes it when the output of `einsum` is a scalar, but it still seems to fail when the output is a `Tensor`.
```nim
import arraymancer

proc einsum_with_scalar_output[T](x, y: Tensor[T]): T =
  result = einsum(x, y):
    x[i, k] * y[i, k]

proc einsum_with_tensor_output[T](x, y: Tensor[T]): Tensor[T] =
  result = einsum(x, y):
    c[i] = x[i, k] * y[i, k]

if is_main_module:
  var x = [[0, 0, 1], [1, 0, 0], [1, 1, 0]].to_tensor.astype(float32)
  var y = [[0, 0, 1], [0, 1, 0], [0, 0, 1]].to_tensor.astype(float32)

  # OK (fixed in #537).
  # Expected: 1.0
  echo einsum_with_scalar_output(x, y)

  # Error: Does not compile.
  # Expected: [1.0 0.0 0.0]
  echo einsum_with_tensor_output(x, y)

  # OK.
  let r = einsum(x, y):
    c[i] = x[i, k] * y[i, k]
  echo r
```
```shell
Hint: used config file '/home/u/.choosenim/toolchains/nim-1.6.2/config/nim.cfg' [Conf]
Hint: used config file '/home/u/.choosenim/toolchains/nim-1.6.2/config/config.nims' [Conf]
.............................................................
/home/u/.nimble/pkgs/nimblas-0.2.2/nimblas/private/common.nim(50, 7) Hint: Using BLAS library with name: libblas.so(||.3|.2|.1|.0) [User]
........................................................................................................................................................................
/home/u/.nimble/pkgs/nimlapack-0.2.0/nimlapack.nim(19, 7) Hint: Using LAPACK library with name: liblapack.so(||.3|.2|.1|.0) [User]
.....................................................................
else xCont[i, k] * yCont[i, k]
stack trace: (most recent call last)
/home/u/.nimble/pkgs/arraymancer-#head/arraymancer/tensor/einsum.nim(561, 32) einsum
/home/u/.nimble/pkgs/arraymancer-#head/arraymancer/tensor/einsum.nim(244, 10) checkStatement
/home/u/src/sandbox/nim-arraymancer/bug_einsum_proc.nim(21, 33) template/generic instantiation of `einsum_with_tensor_output` from here
[...]

PR #544, opened 2021-12-29
Add test for #543

asnt:
Tentative test that should pass when #543 is fixed.


Vindaar:
Commit exists as cherry pick in #545.

PR #546, opened 2021-12-30
Add broadcasted exponentiation scalar with tensor

asnt:
Add `scalar ^. tensor` operation, to complete the already available `tensor ^. scalar`.

For the cpu backend only.


Vindaar:
Thanks a lot!

PR #549, opened 2022-01-20
add stricteffects support

xflywind:
ref https://github.com/nim-lang/Nim/pull/19380

### DIFF
--- a/src/arraymancer/private/functional.nim
+++ b/src/arraymancer/private/functional.nim
@@ -15,6 +15,9 @@
 # Functional programming and iterator tooling
 import sequtils
 
-proc concatMap*[T](s: seq[T], f: proc(ss: T):string): string  {.noSideEffect.}=
+when not defined(nimHasEffectsOf):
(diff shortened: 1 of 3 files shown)


mratsim:
Shouldn't all higher order functions be concerned?

Like

https://github.com/mratsim/Arraymancer/blob/20e0dc371caab1fa6a80edb6e6554e164cdbc238/src/arraymancer/tensor/higher_order_applymap.nim#L149-L171

xflywind:
IMO only procs with effects (`noSideEffect` or `gcsafe`) should be concerned?

xflywind:
Make sense, thanks!

mratsim:
Then I think the last ones are those here: https://github.com/mratsim/Arraymancer/blob/20e0dc3/src/arraymancer/tensor/higher_order_foldreduce.nim#L69-L72

xflywind:
My mistake, thank you!

mratsim (approved):

PR #551, opened 2022-02-10
add ci 1.6

Clonkk:
I think it's time to test arraymance ragainst Nim 1.6 in its CI since 1.6 is a LTS and we're already at the third release of 1.6.X

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,8 @@ jobs:
       fail-fast: false
       max-parallel: 20
       matrix:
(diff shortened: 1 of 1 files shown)


Clonkk:
@mratsim Replacing Nim build with pre-built github action simplify CI and it seems to pass. 

Is there a reason to keep the manual build or does this change seems reasonable to you ?

mratsim (approved):

mratsim:
My manual build predated any pre-built Github Action CI.

Issue #535, opened 2021-11-16
--gc:orc -d:openmp cause a segfault when calling sum(axis = 2)


Clonkk:
```nim
import arraymancer

proc okproc*(tensor1: Tensor[float]) =
  var tmp = tensor1.sum(axis = 2).squeeze
  echo tmp

proc main() =
  let ex1 = ones[float]([20, 20, 8])
  okproc(ex1)

main()
```

Compiled with ``nim cpp --gc:orc -d:openmp`` produce a segfault.

Changin Tensor definition to : 
```nim
(...)
  storage* {.cursor.}: CpuStorage[T]
  ```
  seem to fix the issue

mratsim:
What's the different between SharedPtr and ref in the ORC model?

The way forward is probably to drop OpenMP anyway.

Clonkk:
AFAIK, it's mostly the same thing (syntax differ - SharedPtr behave more like a pointer).

OpenMP offers a good performance boost on apply_inline / map_inline etc. ; what do you have in mind as a replacement? Implementing a backend based on Weave?

PR #555, opened 2022-04-01
Fix memory leak introduced by cursor notation

Clonkk:
* The cursor notation fixed a crash in openmp but caused a memory leak on some cases due to cursor notation on cycle
*  The actual segfault seemed linked to a memory allocation inside an openmp loop ( ``results`` array in ``backend/openmp.nim``).

Many tests later the easiest solution was to disable openmp for omp_parallel_reduce_blocks with ``when defined(gcDestructors)``. Since the goal is to eventually move away from openmp this shouldn't be a problem (an it's better to slow some fold / reduce loop than to cras h I guess ? ) 

@mratsim If you have an actual solution to avoid disabling it entirely I'd be glad to implement it.

### DIFF
--- a/src/arraymancer/laser/tensor/datatypes.nim
+++ b/src/arraymancer/laser/tensor/datatypes.nim
@@ -32,7 +32,7 @@ type
     shape*: Metadata                     # 56 bytes
     strides*: Metadata                   # 56 bytes
     offset*: int                         # 8 bytes
-    storage* {.cursor.}: CpuStorage[T]   # 8 bytes
+    storage*: CpuStorage[T]              # 8 bytes
 
   CpuStorage*[T] {.shallow.} = ref CpuStorageObj[T] # Total heap: 25 bytes = 1 cache-line
   CpuStorageObj[T] {.shallow.} = object
--- a/src/arraymancer/tensor/backend/openmp.nim
+++ b/src/arraymancer/tensor/backend/openmp.nim
@@ -92,7 +92,7 @@ template omp_parallel_reduce_blocks*[T](reduced: T, block_offset, block_size: un
           let num_blocks = min(min(ompsize, omp_get_max_threads()), OMP_MAX_REDUCE_BLOCKS)
           if num_blocks > 1:
             withMemoryOptimHints()
-            var results {.align64.}: array[OMP_MAX_REDUCE_BLOCKS * maxItemsPerCacheLine, type(reduced)]
+            var results {.align64, noInit.}: array[OMP_MAX_REDUCE_BLOCKS * maxItemsPerCacheLine, type(reduced)]
             let bsize = ompsize div num_blocks
 
             if bsize > 1:
--- a/src/arraymancer/tensor/backend/openmp.nim
+++ b/src/arraymancer/tensor/backend/openmp.nim
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import  ./global_config,
+import ./global_config,
         ./memory_optimization_hints
 
 when defined(openmp):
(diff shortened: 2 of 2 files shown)


Clonkk:
This is linked to https://github.com/mratsim/Arraymancer/issues/535 but is actually caused by ``reduce_axis_inline`` which calls ``omp_parallel_reduce_blocks``, so simply :  

```nim
import arraymancer
proc okproc*(tensor1: Tensor[float]): Tensor[float] =
  var tmp0 = tensor1.reduce_axis_inline(2):
    x = 2.0*x + 2.0*y
```

Compiled with ``nim cpp -d:openmp --mm:orc``, this will segfault (note that C backend has an openMP block/structure codegen error that may or may not be related).

The stacktrace obtained with gdb looks like this : 
```
nimDecRefIsLastCyclicStatic (desc=<optimized out>, p=0x7ffff7c01e10 <__pthread_create_2_1+1344>) at /home/rcaillaud/.choosenim/toolchains/nim-#devel/lib/system/orc.nim:495
495	      dec cell.rc, rcIncrement
(gdb) bt
#0  nimDecRefIsLastCyclicStatic (desc=<optimized out>, p=0x7ffff7c01e10 <__pthread_create_2_1+1344>) at /home/rcaillaud/.choosenim/toolchains/nim-#devel/lib/system/orc.nim:495
#1  eqsink___OOZOOZlocalwsZarraymancerZsrcZarraymancerZtensorZmath95functions_1786 (dest=..., src=<optimized out>)
    at /home/rcaillaud/Workspace/localws/arraymancer/src/arraymancer/tensor/higher_order_applymap.nim:80
#2  0x00005555555c36ac in okproc__arrbug_4 (tensor1=<optimized out>, Result=<optimized out>) at /home/rcaillaud/Workspace/localws/arraymancer/src/arraymancer/tensor/higher_order_applymap.nim:80
```
If you add an align pragma to dynamic stack array :
 ```nim
 type DynamicStackArray*[T] = object
    ## Custom stack allocated array that behaves like seq.
    ## We must avoid seq creation when modifying tensor shapes, strides or slicing in a tight loop.
    ## Seq creation are also not possible within an OpenMP loop.
    data* {.align(64).} : array[LASER_MAXRANK, T]
    len*: int
  ```

The stacktrace points to ``results`` array in ``backend/openmp.nim`` the ``omp_parallel_reduce_block`` : 
```
#1  eqdestroy___arrbug_6565 (dest=<optimized out>) at /home/rcaillaud/Workspace/localws/arraymancer/src/arraymancer/tensor/backend/openmp.nim:95
#2  0x00005555555d3f9b in okproc__arrbug_4 (tensor1=<optimized out>, Result=<optimized out>) at /home/rcaillaud/Workspace/localws/arraymancer/src/arraymancer/tensor/backend/openmp.nim:95
#3  0x00005555555d4386 in main__arrbug_6836 () at /home/rcaillaud/Workspace/playground/arraymancerTest/arrbug.nim:27
```


```nim
      when defined(openmp):
        if ompsize * weight >= OMP_FOR_THRESHOLD:
          let num_blocks = min(min(ompsize, omp_get_max_threads()), OMP_MAX_REDUCE_BLOCKS)
          if num_blocks > 1:
            withMemoryOptimHints()
            var results {.align(64), noInit.}: array[OMP_MAX_REDUCE_BLOCKS * maxItemsPerCacheLine, type(reduced)] # This line
```

Clonkk:
@mratsim Do you have any insight on this or is disabling openmp on the fold/reduce proc with``gcDestructors`` the best we can do for now ?

mratsim:
I think it is. I don't have time to look for a better solution unfortunately.

Clonkk:
> I think it is. I don't have time to look for a better solution unfortunately.

No worry, I understand. I think this PR should be merged, it's better to disable openmp than to let a memory leak IMHO. Eventually, this would have to be fixed (possibly by replacing openmp with Weave ;) ) but for now it can wait.

PR #548, opened 2022-01-04
Redesign the neural network macro

tsoj:
```nim
network DemoNet:
  layers h, w:
    cv1:        Conv2D(@[1, h, w], 20, (5, 5))
    mp1:        Maxpool2D(cv1.out_shape, (2,2), (0,0), (2,2))
    cv2:        Conv2D(mp1.out_shape, 50, (5, 5))
    mp2:        MaxPool2D(cv2.out_shape, (2,2), (0,0), (2,2))
    fl:         Flatten(mp2.out_shape)
    hidden:     Linear(fl.out_shape[0], 500)
    classifier: Linear(500, 10)
  forward x:
    x.cv1.relu.mp1.cv2.relu.mp2.fl.hidden.relu.classifier
```

creates now something like this
<details>
  <summary>Click to expand!</summary>

```nim
type
  DemoNet[T] = object
    cv1: Conv2D[T]
    mp1: Maxpool2D[T]
    cv2: Conv2D[T]
    mp2: MaxPool2D[T]
    fl: Flatten[T]
    hidden: Linear[T]
    classifier: Linear[T]

proc init[T](ctx: Context[AnyTensor[T]]; model_type: typedesc[DemoNet[T]]; h: auto; w: auto): DemoNet[T] =
  template cv1(): auto =
    result.cv1

  template mp1(): auto =
    result.mp1

  template cv2(): auto =
    result.cv2

  template mp2(): auto =
    result.mp2

  template fl(): auto =
    result.fl

  template hidden(): auto =
    result.hidden

  template classifier(): auto =
    result.classifier

  cv1 = init(ctx, Conv2D[T], @[1, h, w], 20, (5, 5))
  mp1 = init(ctx, Maxpool2D[T], cv1.out_shape, (2, 2), (0, 0), (2, 2))
  cv2 = init(ctx, Conv2D[T], mp1.out_shape, 50, (5, 5))
  mp2 = init(ctx, MaxPool2D[T], cv2.out_shape, (2, 2), (0, 0), (2, 2))
  fl = init(ctx, Flatten[T], mp2.out_shape)
  hidden = init(ctx, Linear[T], fl.out_shape[0], 500)
  classifier = init(ctx, Linear[T], 500, 10)

proc forward[T](self: DemoNet[T]; x: auto): auto =
  template cv1(x: varargs[untyped]): auto =
    forward(self.cv1, x)

  template mp1(x: varargs[untyped]): auto =
    forward(self.mp1, x)

[...]

PR #556, opened 2022-05-07
Update docs, add GH action for docs, fix a few styleCheck hints

Vindaar:
Let's try to build the docs on GH actions...

Fixes the doc gen for the current behavior of `nim doc` as well as update it for PR #548.


Vindaar:
Let's merge this and see if the docs actually get built...

PR #557, opened 2022-05-11
Refactor `optimize*` functions to use generics

tandy-1000:
This was in the TODOs, and it seems to work.
I have also changed Sgd* to SGD for consistency, not sure if that is wanted though.

### DIFF
--- a/src/arraymancer/nn/optimizers/optimizers.nim
+++ b/src/arraymancer/nn/optimizers/optimizers.nim
@@ -150,8 +150,8 @@ proc optimizer*[M, T](
         model: M,
         OptimizerKind: typedesc[SGDMomentum],
         learning_rate: T,
-        momentum = T(0.0),
-        decay = T(0.0),
+        momentum: T = T(0.0),
+        decay: T = T(0.0),
         nesterov = false): SGDMomentum[Tensor[T]] =
   ## Create a SGD optimizer with optional momentum that will update the model weight
   ##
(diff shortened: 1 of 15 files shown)


mratsim (commented):
We need a deprecation path to avoid breaking everyone's code. Otherwise perfect, thanks

tsoj:
Is there an intent behind using `optimize` instead of `optimizer`? The former one could be confusing because the function doesn't optimize, it only provides an object that can optimize.

mratsim:
Since I was writing optimizerSGD / optimizerAdam / optimizerSGDMomentum, I assume I also thought the same. And the `optimize` was just a shortname I used when trying to debug and reproduce the bug.

So yes `optimizer` is the better name I agree. Can you replace all @tandy-1000, sorry for that update.

mratsim:
> Does [nim-lang/Nim#7733](https://github.com/nim-lang/Nim/issues/7733) need to be closed now?

I'm not too sure why but on 1.6.6 #7733 still doesn't work

tandy-1000:
Done, also fixed a bug.

src/arraymancer/nn/optimizers/optimizers.nim
```diff
@@ -159,20 +164,17 @@ func optimizerSGDMomentum*[M, T](model: M, learning_rate: T, momentum = T(0.0),
...
-  # Pending https://github.com/nim-lang/Nim/issues/7734 and https://github.com/nim-lang/Nim/issues/7733
 
   result.params = @[]
   result.lr = learning_rate
   result.momentum = momentum
   result.decay = decay
   result.nesterov = nesterov
 
-  func addLayer(result: var SgdMomentum[Tensor[T]], layer: auto) =
+proc optimizerSGDMomentum*[M, T](model: M, learning_rate: T, momentum = T(0.0), decay = T(0.0), nesterov = false): SGDMomentum[Tensor[T]] {.deprecated: "Use optimizer(model, SGDMomentum, learning_rate) instead."} =
```
mratsim, line 174: The addLayer part should be in the `optimizer` proc no?

tandy-1000: oops, fixed.

Vindaar:
Should this be merged?

PR #558, opened 2022-05-12
Conv2D Docs correction

tandy-1000:
Since `inShape` must be len 3 it should say "- inShape Shape of the expected input tensor in the form of [C_in, H_in, W_in]".


tandy-1000:
Removed the merge conflicts, this can be squash merged now

PR #559, opened 2022-05-22
returnEmptyIfEmpty returns tensor of size 0

paulnorrie:
The function `returnEmptyIfEmpty` previously returned an unitialized tensor of size 1.  

This occured because if would call, for example, `Tensor[int]()` instead of `newTensor[int](0)`.  Calling `Tensor[int]()` results in an unitialized tensor that a call to `isEmpty` fails and a call to `size` results in `1`.


mratsim:
Thank you!

Can you a test for the correct behavior?

paulnorrie:
Sorry, that was a fail on my part.  It now compiles and I'm testing the public interface with a test calling `astype`

PR #563, opened 2022-07-25
add CI for ORC and devel

Vindaar:
Attempt to also test ORC in the CI as well as Nim devel.

I run the tests for ORC in release mode to not slow down the full CI too much.


Vindaar:
ORC release tests fail, I think, because of:

https://github.com/nim-lang/Nim/issues/19374

:(

PR #564, opened 2022-07-28
Quiet hints about unused imports, style checks etc.

Vindaar:
Pheew, this was some tedious work.

I added the `styles:usages` flag to arraymancer's `nim.cfg` and then went for a hunt for variables / procs etc. that violated our rules of naming. In almost all cases I simply stuck to the name initially used.

Personally I would probably change `astype` to `asType`. But given that it's `noinit` and `nodecl` instead of `noInit` and `noDecl`, I guess `astype` also makes sense...

I'd appreciate if someone could give some input on this, as I had to touch a big chunk of the code.

One note: for some templates that used a `t` Tensor argument and a `T` generic type, the compiler got confused thinking small `t` referred to type `T`. So in those cases I changed the argument name to `x` or `arg`.


HugoGranstrom:
I agree with you on `asType` and all names that we define in our libraries should be camelCase. The fact that `noinit` isn't, is not our concern really. I think it's better to keep that kind of style only the the compiler-specific ones because those people are more likely to know about.

mratsim:
`asType`  is fine

c-blake:
Personally, I think the compiler should change `noinit`/`nodecl` to `noInit`/`noDecl` and in general do a clean-up - sooner is better than later since people will probably only use style-check more as time goes on.

Clonkk:
@mratsim Considering CI 1.0 and 1.2 have been dropped by commit a59d703d1d4ee6371e9e9c679c711ec02cfea4b0 I think we should be able to merge this

PR #565, opened 2022-08-04
misc k-d tree fixes & fixes `deepCopy` for tensor

Vindaar:
This is based on #564. I'll rebase it onto master once that is merged.

It's just a few small fixes to the k-d tree module. The `TensorHelper` hacky type is removed and replaced by a custom `<` proc that is bound in the generic scope.

Further it fixes the `deepCopy` implementation of `Tensor`, as that was outdated I think.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,11 @@
 Arraymancer v0.7.x
 =====================================================
 
+- add `clone` operation for k-d tree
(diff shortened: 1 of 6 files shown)


Clonkk:
@mratsim Considering CI 1.0 and 1.2 have been dropped by commit a59d703d1d4ee6371e9e9c679c711ec02cfea4b0 I think we should be able to merge this

PR #567, opened 2022-08-16
Improve the display of rank 2 column Tensors

AngelEzquerra:
This also adds a little more documentation to the Flatten NN layer.


Vindaar:
Sorry, I overlooked the actual PR! Could you maybe add a single test case for the display of a simple transposed 1D tensor?

AngelEzquerra:
> Sorry, I overlooked the actual PR! Could you maybe add a single test case for the display of a simple transposed 1D tensor?

That is a good idea. I just pushed an additional commit which adds a new display test.

Vindaar:
You accidentally also commited the `test_display` binary.

Vindaar:
Thanks a lot!

PR #571, opened 2022-09-07
remove the overload of slicer with multiple varargs 

ringabout:
multiple varargs never worked well, even a simple case breaks

```nim
proc thing(a: varargs[string], m: int, b: varargs[float]) = 
  discard

thing("1.0", "2.0", 0, 1.0) # compiles
thing("1.0", 0, 1.0, 2.0) # broken
thing("1.0", "2.0", 0, 1.0, 2.0) # broken
```

It is about to become a compilation error => https://github.com/nim-lang/Nim/pull/20266

 `slicer` seems to be a function in a private module and not used internally. I suppose it can be safely modified.

### DIFF
--- a/src/arraymancer/tensor/private/p_accessors_macros_read.nim
+++ b/src/arraymancer/tensor/private/p_accessors_macros_read.nim
@@ -101,7 +101,7 @@ proc slicer*[T](t: AnyTensor[T],
 
   result = t
   let full_slices = concat(slices1.toArrayOfSlices,
-                            initSpanSlices(t.rank - slices1.len - slices2.len),
+                            initSpanSlices(t.rank - 1 - 1),
                             slices2.toArrayOfSlices)
   slicerImpl(result, full_slices)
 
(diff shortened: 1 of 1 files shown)


Vindaar:
This is used internally to resolve the `[]` macro logic. Fixing it will probably require going into the macro logic that deconstructs the inputs and generates the `slicer` call.

ringabout:
You are right, I will work on a solution.

mratsim:
The easiest would likely be to have all the slicer and slicer_mut be an openarray.

The macros that do the dispatching can repack them into a fixed size array (the actual number of args is always known at compile-time):

1. So either at the untyped dispatch level: https://github.com/mratsim/Arraymancer/blob/66372d249b849102fd196345a08d4563d904751a/src/arraymancer/tensor/accessors_macros_read.nim#L20-L56
    and https://github.com/mratsim/Arraymancer/blob/66372d249b849102fd196345a08d4563d904751a/src/arraymancer/tensor/accessors_macros_write.nim#L21-L52
2. Alternatively at the typed dispatch level 
   https://github.com/mratsim/Arraymancer/blob/66372d249b849102fd196345a08d4563d904751a/src/arraymancer/tensor/private/p_accessors_macros_read.nim#L196-L244
   and https://github.com/mratsim/Arraymancer/blob/66372d249b849102fd196345a08d4563d904751a/src/arraymancer/tensor/private/p_accessors_macros_write.nim#L321-L338

ringabout:
succeeded by https://github.com/mratsim/Arraymancer/pull/572

PR #573, opened 2022-09-14
make arraymancer compatible with ORC and latest devel 

ringabout:
Closes https://github.com/mratsim/Arraymancer/pull/563
ref https://github.com/nim-lang/Nim/pull/19972
ref https://github.com/nim-lang/Nim/issues/19374


mratsim (approved):
Thanks

PR #575, opened 2022-11-01
Fix #574

### LINKED ISSUE #574: set[int] used that depends on odd compiler behavior.
https://github.com/mratsim/Arraymancer/blob/276cacd2503d4f3e3270616eafd80d5b0625fa9d/src/arraymancer/nn/init.nim#L52

bung87:
Fix #574


Vindaar:
Thanks and sorry for the delay!

PR #578, opened 2022-12-02
Workaround upstream issue and change `KnownSupportsCopyMem`

Vindaar:
The explicit list of types supporting `copyMem` has always been a band-aid fix. It make it impossible to use a `Tensor` with a mem copyable custom type (e.g. a `distinct float`). By changing `KnownSupportsCopyMem` to a concept we can use the real `supportsCopyMem` check, while still using it in a type section to define the `CpuStorage` type.

Further fixes an old bug fix workaround, due to the need to fix the upstream issue 
https://github.com/nim-lang/Nim/issues/20993
which fortunately for us can be done by simplifying the `toTensor` logic.

### DIFF
--- a/src/arraymancer/laser/tensor/datatypes.nim
+++ b/src/arraymancer/laser/tensor/datatypes.nim
@@ -17,7 +17,9 @@ when not defined(nimHasCursor):
   {.pragma: cursor.}
 
 type
-  KnownSupportsCopyMem* = SomeNumber | char | Complex[float64] | Complex[float32] | bool
+  ## This works around the issue that `supportsCopyMem` is not supported in a type section
+  KnownSupportsCopyMem* = concept x, type T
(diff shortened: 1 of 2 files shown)


Vindaar:
The CI failure seems to be unrelated. Even without the changes (so on master) on my local machine using latest devel we get the same recursive call. Did something change about the type matching for `int` vs. `int64`? Or rather does the current code depend on the compiler matching `int` and `int64` differently?

Vindaar:
This has broken:

```nim
echo int64 is (SomeNumber and not(uint32|uint64|uint|int))
```

should be `true`, but is `false` now.

See: https://play.nim-lang.org/#ix=4hrM

Vindaar:
Merging anyway, as the CI failure is unrelated and already present on `master`.

PR #588, opened 2023-06-12
fixes math path

ringabout:
ref https://github.com/nim-lang/Nim/pull/22073


Vindaar:
Thanks a lot. :)

Vindaar:
Sigh, I'll take a look at that SVD failure.

ringabout:
I have downgrade Ubuntu 22.04 to 20.04. Guess there is something wrong with newer GCC or .so library,

PR #591, opened 2023-09-25
rename `newSeqUninit` in order to avoid conflicts with upstream

ringabout:
ref https://github.com/nim-lang/Nim/pull/22739

It's a private function so it shouldn't affect backwards compatibility.

### DIFF
--- a/src/arraymancer/autograd/autograd_common.nim
+++ b/src/arraymancer/autograd/autograd_common.nim
@@ -241,7 +241,7 @@ proc backprop*[TT](v: Variable[TT]) =
         parent_i.grad += diff
 
 func newParents*[TT](num: Natural): Parents[TT] {.inline.} =
-  newSeqUninit[Variable[TT]](num)
+  newSeqUninit2[Variable[TT]](num)
 
 func newDiffs*[TT](num: Natural): SmallDiffs[TT] {.inline.} =
-  newSeqUninit[TT](num)
+  newSeqUninit2[TT](num)
--- a/src/arraymancer/autograd/gates_shapeshifting_concat_split.nim
+++ b/src/arraymancer/autograd/gates_shapeshifting_concat_split.nim
@@ -67,7 +67,7 @@ proc stack*[TT](variables: varargs[Variable[TT]], axis = 0): Variable[TT] =
 
   # Resulting var
   new result
-  var ts = newSeqUninit[TT](variables.len)
+  var ts = newSeqUninit2[TT](variables.len)
   for i in 0 ..< variables.len:
     # TODO: avoid the intermediate seq alloc to extract varargs Tensors from varargs variables
     ts[i] = variables[i].value
--- a/src/arraymancer/linear_algebra/decomposition.nim
+++ b/src/arraymancer/linear_algebra/decomposition.nim
@@ -80,7 +80,7 @@ proc qr*[T: SomeFloat](a: Tensor[T]): tuple[Q, R: Tensor[T]] =
   let k = min(a.shape[0], a.shape[1])
 
   var scratchspace: seq[T]
-  var tau = newSeqUninit[T](k)
+  var tau = newSeqUninit2[T](k)
 
   result.Q = a.clone(colMajor)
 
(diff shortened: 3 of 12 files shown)


Vindaar:
I'm confused as to why we cannot keep the `when not declared` logic? I haven't fully understood the stdlib changes yet, sorry.

ringabout:
It seems that https://github.com/mratsim/Arraymancer/blob/875ee6bf595a220a08827adeaee8a89722d2c8f2/src/arraymancer/laser/tensor/datatypes.nim#L37 uses refs, which is not supported by `newSeqUninit`.

But `newSeqUninit` is unavailable to types, which contain managed memory or have destructors.

Vindaar:
Ohh, now I see. My confusion was that I assumed this PR here https://github.com/nim-lang/Nim/pull/22586 was merged already and based on Mamy's PR #591 I further assumed it was a drop in replacement for our custom one. 

And yes, the `CpuStorage` can either hold regular POD types or references. Thanks for the explanation.

What is the reason though for the restriction of the stdlib version to types that can be memcopied? The code currently doesn't do anything that seems to rely on that, no? 

I don't have a strong opinion on the name, `newSeqUninit2` is fine by me. Just trying to understand the restrictions.

ringabout:
I think type bounding operations requires types to be initialized preemptively.

For instance, destructors or other type bounding operations might not work properly since `Ref` might be anything.
```nim
var x = newSeqNoInit[Ref](10)
```

Vindaar:
But then what does our current code do? Or in other words, if I currently write:
```nim
type
  Foo = ref object
    x: float

var x = newSeqOfCap[Foo](10)
x.setLen(10)
```
wouldn't that be undefined behavior if I understand you correctly?

ringabout:
`setLen` will initialize the sequence in v2 since its original length is smaller than the new length
```nim
var x = newSeqOfCap[Foo](10)
x.setLen(10)
```
so it is just another way of writing `vat x = newSeq[Foo](10)` with the same efficiency.

Vindaar:
Ah, I see. But then that implies our implementation, `newSeqUninit2` now, is "broken" and doesn't actually do an "uninit" implementation anymore. 

So we need to merge the logic for `newSeqUninit2` to be something like this:

```nim
func newSeqUninit2*[T](len: Natural): seq[T] {.inline.} =
  ## Creates an uninitialzed seq if the type is plain old data. Falls back
  ## to a normal `newSeq` call for any non trivial data type.
  when supportsCopyMem(T):
    result = newSeqUninit[T](len)
  else:
    result = newSeq[T](len)
```

so that at least we get the desired behavior for mem copyable types.

ringabout:
Indeed, I agree. It also need to be combined with `when declared(newSeqUninit)`.

Vindaar:
Cool! Will you make the change or should I do it?

ringabout:
Feel free to take over this PR, or do it in the following PRs. Thank you in advance!

Vindaar:
Sorry for the delay. Will take care of this soon. Yesterday was a bit stressful. :)

mratsim:
I think we can just delete newSeqUninit in Arraymancer actually, with the Laser PR, the storage for POD datatypes changed to ptr UncheckedArray:

https://github.com/mratsim/Arraymancer/blob/875ee6bf595a220a08827adeaee8a89722d2c8f2/src/arraymancer/laser/tensor/datatypes.nim#L39C1-L47

so CPU storage is only used for ref types and strings but the current impl will likely cause issues with destructors and we aren't optimizing for strings/ref types.

In fact, CpuStorage uses `newSeq` for alloc of tensors

https://github.com/mratsim/Arraymancer/blob/875ee6bf595a220a08827adeaee8a89722d2c8f2/src/arraymancer/laser/tensor/datatypes.nim#L86-L101

and newSeqUninit only for intermediate sequences when interfaces with C libraries like Lapack

https://github.com/search?q=repo%3Amratsim%2FArraymancer%20newSequninit&type=code

Vindaar:
Good point actually.
The one exception is the `toRawSeq` implementation, but given that this is deprecated for good reason, I'd simply use a regular `newSeq` there instead and go ahead and removing our implementation.

PR #592, opened 2023-09-28
Remove custom `newSeqUninit` if available in stdlib

Vindaar:
Takes over from #591. We now use the stdlib `newSeqUninit` if available, i.e. after https://github.com/nim-lang/Nim/pull/22739.

### DIFF
--- a/src/arraymancer/autograd/autograd_common.nim
+++ b/src/arraymancer/autograd/autograd_common.nim
@@ -241,7 +241,7 @@ proc backprop*[TT](v: Variable[TT]) =
         parent_i.grad += diff
 
 func newParents*[TT](num: Natural): Parents[TT] {.inline.} =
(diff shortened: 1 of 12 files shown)


ringabout:
Hi, it doesn't seem to handle all the cases.

> /home/runner/work/Nim/Nim/pkgstemp/arraymancer/src/arraymancer/autograd/autograd_common.nim(247, 19) template/generic instantiation of `newSeqUninit` from here
  /home/runner/work/Nim/Nim/lib/system.nim(1659, 14) Error: The type T cannot contain managed memory or have destructors

ref https://github.com/nim-lang/Nim/pull/22782

Vindaar:
Oh, I didn't realize these would be non mem copyable. I'll investigate and fix it!

PR #594, opened 2023-10-05
Add support for decomposition complex tensors, to calculate the pseudo-inverse and the conjugate of a complex tensor

AngelEzquerra:
A lot of the code on this PR was actually written by @Vindaar, who provided the code to support the decomposition of complex tensors, to which I added a test case. In addition to that I have added pinv functions to calculate the pseudo-inverse (which are based on the svd decomposition functions) and made some of the common error functions support complex tensors.

Note that I made the pinv API match the corresponding numpy.linalg.pinv function API and in fact I copied most of the description from the numpy docs.


src/arraymancer/linear_algebra/algebra.nim
```diff
@@ -0,0 +1,54 @@
...
+  ##            Singular values less than or equal to
+  ##            `rcond * largest_singular_value` are set to zero.
+  var (U, S, Vt) = A.svd()
+  let epsilon = S.max() * rcond
+  let S_size = S.shape[0]
+  var S_inv = zeros[T]([S_size, S_size])
+  const unit = T(1.0)
+  for n in 0..<S.shape[0]:
+    S_inv[n, n] = if abs(S[n]) > epsilon: unit / S[n] else: 0.0
+  Vt.transpose * S_inv * U.transpose
```
Vindaar, line 33:
```suggestion
  result = Vt.transpose * S_inv * U.transpose
```

src/arraymancer/linear_algebra/algebra.nim
```diff
@@ -0,0 +1,54 @@
...
+  ##            Singular values less than or equal to
+  ##            `rcond * largest_singular_value` are set to zero.
+  var (U, S, Vt) = A.svd()
+  let epsilon = S.max() * rcond
+  let S_size = S.shape[0]
+  var S_inv = zeros[T]([S_size, S_size])
+  const unit = complex(1.0)
+  for n in 0..<S.shape[0]:
+    S_inv[n, n] = if abs(S[n]) > epsilon: unit / S[n] else: complex(0.0)
+  Vt.conjugate.transpose * S_inv * U.conjugate.transpose
```
Vindaar, line 54:
```suggestion
  result = Vt.conjugate.transpose * S_inv * U.conjugate.transpose
```

Vindaar (commented):
Outside of the CI failure (requiring a `address` helper that uses `unsafeAddr` for older Nim) this is great! The `result` changes are not strictly needed of course.

AngelEzquerra:
> Outside of the CI failure (requiring a `address` helper that uses `unsafeAddr` for older Nim) this is great! The `result` changes are not strictly needed of course.

I've fixed the CI failure on the latest commit as you suggested.
Regarding the explicit result assignment, I did not do it because I saw other procedures in the library that didn't do it either and because recently Araq adviced me to use this style for short procedures. However, if you think it is best to change it I will.

Vindaar:
Thanks a lot! :partying_face:

PR #595, opened 2023-10-07
Warning and Hint removal

AngelEzquerra:
Remove a few warnings and hints. There are still a few left. Those are mostly related to =destroy and some unused variables which I think make the code clearer so I prefer not to fix them for now.


Vindaar:
Thanks again!

PR #597, opened 2023-10-11
Add procedures to create and handle diagonal matrices

AngelEzquerra:
Add procedures to create diagonal matrices as well as to set and get diagonals from matrices.

### DIFF
--- a/src/arraymancer/linear_algebra/special_matrices.nim
+++ b/src/arraymancer/linear_algebra/special_matrices.nim
@@ -163,6 +163,7 @@ proc identity*[T](n: int): Tensor[T] {.noInit.}=
   ## Return an identity matrix (i.e. 2-D tensor) of size `n`
   ## 
   ## The identity matrix is a square 2-D tensor with ones on the main diagonal and zeros elsewhere.
+  ## This is basically the same as calling `eye(n, n)`.
   ##
   ## Input:
   ##      - Number of rows / columns in the output.
@@ -170,13 +171,11 @@ proc identity*[T](n: int): Tensor[T] {.noInit.}=
   ##      - The constructed indentity 2-D tensor
   result = diag(ones[T](n))
 
-proc eye*[T](shape: varargs[int]): Tensor[T] {.noInit.}=
+proc eye*[T](shape: varargs[int]): Tensor[T] {.noInit.} =
   ## Return a 2-D tensor with ones on the diagonal and zeros elsewhere
   ##
   ## Input:
-  ##      - Rank-1 tensor containg the elements of the diagonal
-  ##      - The index of the diagonal that will be set. The default is 0.
-  ##        Use k>0 for diagonals above the main diagonal, and k<0 for diagonals below the main diagonal.
+  ##      - The shape of the output matrix
   ## Result:
   ##      - The constructed, rank-2 diagonal tensor
   result = zeros[T](shape)
--- a/src/arraymancer/linear_algebra/special_matrices.nim
+++ b/src/arraymancer/linear_algebra/special_matrices.nim
@@ -58,7 +58,7 @@ proc diagonal*[T](a: Tensor[T], k = 0, anti = false): Tensor[T] {.noInit.} =
   ##      - anti: If true, get the k-th "anti-diagonal" instead of the k-th regular diagonal.
   ## Result:
   ##      - A copy of the diagonal elements as a rank-1 tensor
-  assert a.rank == 2
+  assert a.rank == 2, "diagonal() only works on matrices"
   assert k < a.shape[0], &"Diagonal index ({k=}) exceeds matrix height ({a.shape[0]})"
   assert k < a.shape[1], &"Diagonal index ({k=}) exceeds matrix width ({a.shape[1]})"
   let size = min(a.shape[0], a.shape[1]) - abs(k)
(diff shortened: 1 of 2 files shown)


src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -47,3 +47,137 @@ proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
   let orders = arange(order.float)
   for i, ax in enumerateAxis(result, axis = 1):
     result[_, i] = (x ^. orders[i]).unsqueeze(axis = 1)
+
+proc diag*[T](d: Tensor[T], k = 0): Tensor[T] {.noInit.}=
```
Vindaar, line 51:
```suggestion
proc diag*[T](d: Tensor[T], k = 0): Tensor[T] {.noInit.} =
```

src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -47,3 +47,137 @@ proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
...
+  result = zeros[T](size, size)
+
+  if k >= 0:
+    for i in 0 ..< d.size:
+      result[i, i+k] = d[i]
+  else:
+    for i in 0 ..< d.size:
+      result[i-k, i] = d[i]
+
+proc diagonal*[T](a: Tensor[T], k = 0, anti = false): Tensor[T] {.noInit.}=
```
Vindaar, line 71:
```suggestion
proc diagonal*[T](a: Tensor[T], k = 0, anti = false): Tensor[T] {.noInit.} =
```

src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -47,3 +47,137 @@ proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
...
+  ##      - anti: If true, set the k-th "anti-diagonal" instead of the k-th regular diagonal.
+  assert a.rank == 2
+  assert d.rank == 1
+  assert k < a.shape[0], &"Diagonal index ({k=}) exceeds matrix height ({a.shape[0]})"
+  assert k < a.shape[1], &"Diagonal index ({k=}) exceeds matrix width ({a.shape[1]})"
+  if anti:
+    if k >= 0:
+      when compileOption("boundChecks"):
+        let size = min(a.shape[0] - abs(k), a.shape[1])
+        assert size == d.size, &"Input size ({d.size}) does not match the {k}-th upper anti-diagonal size ({size})"
```
Vindaar, line 127: I'm not sure if this is what we would "intend". For example if the user wishes bound checks on a `danger` build the assertion is still going to be ignored. If we explicitly activate the assertion only for bounds checks being active, then maybe we should use `doAssert` to also have it work on `danger` builds or use a regular exception? Same for the other `when` below of course.

src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -47,3 +47,137 @@ proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
...
+      for i in 0 ..< d.size:
+        a[i, i+k] = d[i]
+    else:
+      when compileOption("boundChecks"):
+        let size = min(a.shape[0] - abs(k), a.shape[1])
+        assert size == d.size, &"Input size ({d.size}) does not match the {-k}-th lower diagonal size ({size})"
+      for i in 0 ..< d.size:
+        a[i-k, i] = d[i]
+
+proc with_diagonal*[T](a: Tensor[T], d: Tensor[T], k = 0, anti = false): Tensor[T] {.noInit.}=
```
Vindaar, line 150:
```suggestion
proc with_diagonal*[T](a: Tensor[T], d: Tensor[T], k = 0, anti = false): Tensor[T] {.noInit.} =
```

src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -47,3 +47,137 @@ proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
...
+  ## Input:
+  ##      - The matrix that will copied into the output.
+  ##      - Rank-1 tensor containg the elements that will be copied into the selected diagonal.
+  ##      - k: The index k of the diagonal that will be changed. The default is 0 (i.e. the main diagonal).
+  ##        Use k>0 for diagonals above the main diagonal, and k<0 for diagonals below the main diagonal.
+  ##      - anti: If true, set the k-th "anti-diagonal" instead of the k-th regular diagonal.
+  result = a
+  set_diagonal(result, d, k, anti=anti)
+
+proc identity*[T](n: int): Tensor[T] {.noInit.}=
```
Vindaar, line 162:
```suggestion
proc identity*[T](n: int): Tensor[T] {.noInit.} =
```

src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -47,3 +47,137 @@ proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
...
+  ## 
+  ## The identity matrix is a square 2-D tensor with ones on the main diagonal and zeros elsewhere.
+  ##
+  ## Input:
+  ##      - Number of rows / columns in the output.
+  ## Result:
+  ##      - The constructed indentity 2-D tensor
+  result = diag(ones[T](n))
+
+proc eye*[T](shape: varargs[int]): Tensor[T] {.noInit.}=
```
Vindaar, line 173:
```suggestion
proc eye*[T](shape: varargs[int]): Tensor[T] {.noInit.} =
```

src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -47,3 +47,137 @@ proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
...
+  ## Return a 2-D tensor with ones on the diagonal and zeros elsewhere
+  ##
+  ## Input:
+  ##      - Rank-1 tensor containg the elements of the diagonal
+  ##      - The index of the diagonal that will be set. The default is 0.
+  ##        Use k>0 for diagonals above the main diagonal, and k<0 for diagonals below the main diagonal.
```
Vindaar, line 179: The doc string seems wrong. And do I understand correctly that the difference between `eye` and `identity` is just that it allows to construct rectangular matrices? What happens with `shape.len > 2`?

Vindaar (changes requested):
Generally great work! A few minor things as noted in the comments.

And another one: Shouldn't we implement `diag` using `set_diagonal` as well? Then `diag` can also support `anti` and we remove some duplicated code.

src/arraymancer/linear_algebra/special_matrices.nim
AngelEzquerra: You are right, implementing the diag procedure based on set_diagonal makes more sense as it reduces code duplication and let's us easily add support for the anti flag as well.
I will fix the missing spaces as well.

src/arraymancer/linear_algebra/special_matrices.nim
AngelEzquerra: I will fix this.

src/arraymancer/linear_algebra/special_matrices.nim
AngelEzquerra: I will replace the calls to assert with doAssert as you suggest.

AngelEzquerra:
The new changes should hopefully address all the existing comments.

Vindaar (approved):
Looks great now!

PR #598, opened 2023-10-11
Add support for combining span slices with negative steps.

AngelEzquerra:
This adds support for `a[_|-2]` and `a[_.._|-2]` syntax, both of which are syntactic sugar for `a[^1..0|-2]`. That is, we assume that the intention behind a "full" span combined with a negative step is to go through the tensor in reverse order.

This makes is extremelly simple to reverse a tensor (i.e. `a[_|-1]`) and is equivalent to python's `a[::-1]` syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Arraymancer tutorial is available [here](https://mratsim.github.io/Arraymancer/t
 Here is a preview of Arraymancer syntax.
 
 ### Tensor creation and slicing
+
 ```Nim
 import math, arraymancer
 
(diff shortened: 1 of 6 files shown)


Vindaar:
Nice! Previously these would just throw CT errors (or RT errors?) anyway, right?

Vindaar:
Ah, and could you add a couple of examples here:

https://github.com/mratsim/Arraymancer/blob/master/docs/tuto.slicing.rst

? 

That way (once I fix the doc generation, sorry :disappointed: ) they appear in the docs here https://mratsim.github.io/Arraymancer/tuto.slicing.html

AngelEzquerra:
> Nice! Previously these would just throw CT errors (or RT errors?) anyway, right?

These would throw RT errors. For example:

```nim
echo foo[_:-1]
```
> Error: unhandled exception: Your slice start: 0, and stop: 4, or your step: -1, are not correct. If your step is positive
>                 start must be inferior to stop and inversely if your step is negative
>                 start must be superior to stop. [IndexDefect]

README.md
```diff
@@ -53,11 +52,9 @@ const
     y = @[1, 2, 3, 4, 5]
 
 var
-    vandermonde: seq[seq[int]]
+    vandermonde: newSeq[seq[int]]()
```
Vindaar, line 55: If you want to change it to `newSeq` you forgot to replace the `:` by a `=`

Vindaar:
Great work! Especially also on the README and docs side. Much more than I had in mind! :smile:

README.md
AngelEzquerra: Good point. Sorry for the mistake. I just fixed it.

Vindaar:
Thanks a lot again!

PR #599, opened 2023-10-17
Complex Tensor improvements

AngelEzquerra:
This PR adds functions to create Complex Tensors from real tensors, to get the real and imaginary components of Complex Tensors, and adds a phase procedure to get the phase of a Complex Tensor.

This also moves the complex.nim module from linear_algebra to tensor, moves the conjugate function from the algebra.nim module to complex.nim and adds a few tests for the new operations.


Vindaar:
And again thanks a lot! That's really neat.

I think you still need to add the `test_complex` test case here:

https://github.com/mratsim/Arraymancer/blob/master/tests/tests_cpu.nim

so that it is actually being run as part of `test_cpu`.

PR #600, opened 2023-10-17
Add "item" function to convert a Tensor of size 1 (and any rank) into a scalar value

AngelEzquerra:
This can be used, for example, to extract the result of the dot product of a row and a column Tensor.


Vindaar:
Thank you!

PR #603, opened 2023-10-19
Operator_enhancements

AngelEzquerra:
This PR adds several operator enhancements:

1. It makes Tensor-Scalar additions and subtractions behave in a way that is more consistent with Matlab and numpy.

In particular, up until now you could not use +, -, += and -= with a scalar, you had to use the "dot" versions of those operators. This was inconsistent with the * operation, which can _only_ be used without a dot. It was also inconsistent with how Matlab and numpy works (in both cases you can add or subtract a scalar without a dot.

2. Adds broadcast support to the mod operator

In addition to this we might want to add support for using *. with a scalar, which is currently not supported, while *.= is supported and +. is supported.

### DIFF
--- a/docs/howto.type_conversion.rst
+++ b/docs/howto.type_conversion.rst
@@ -2,7 +2,7 @@
 Spellbook: How to convert a tensor underlying type?
 ===================================================
 
-A type conversion fonction ``asType`` is provided for convenience
+A type conversion function ``asType`` is provided for convenience
 
 .. code:: nim
 
--- a/src/arraymancer/tensor/operators_blas_l1.nim
+++ b/src/arraymancer/tensor/operators_blas_l1.nim
@@ -64,6 +64,26 @@ proc `-=`*[T: SomeNumber|Complex[float32]|Complex[float64]](a: var Tensor[T], b:
 # #########################################################
 # # Tensor-scalar linear algebra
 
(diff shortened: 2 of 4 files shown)


AngelEzquerra:
I've added the .* version of the * operator, an integer version of the Tensor / Scalar operator and a number of tests.

mratsim:
It's an intentional design choice to have a dot prefix for broadcasting operations.

Broadcasting is an ergonomic convention but departs from mathematics.

This is a choice that Julia did as well.

The `*` doesn't need it because `a*X` is a valid mathematical notation when a is a Scalar and X a vector or matrix.

AngelEzquerra:
> It's an intentional design choice to have a dot prefix for broadcasting operations.
> 
> Broadcasting is an ergonomic convention but departs from mathematics.
> 
> This is a choice that Julia did as well.
> 
> The `*` doesn't need it because `a*X` is a valid mathematical notation when a is a Scalar and X a vector or matrix.

I understand and I agree with that. I think broadcasting operations should use the "dot" notation. However, all operations that make sense mathematically should work without a dot (or at least not some and not others, which is super confusing). That is why I added "non dot" operations for vector +/- scalar (and vice-versa), but not for vector +/- vector. I also added for tensor / scalar but not scalar / tensor since 1 / tensor is not the same as 1 / divided by every element.

I hope that makes sense.

AngelEzquerra:
After several conversations, I am closing this pull request and make a new one that takes into account @mratsim's comments.

PR #608, opened 2023-10-21
Add bounds checks for fold and reduce axis inline operations

AngelEzquerra:
The errors that you got when you tried to reduce or fold a tensor along a dimension that does not exist were pretty cryptic. Hopefully these will be an improvement.

It would be great if you could somehow refer to the "op" that is used for the reduction or the fold but I did not find a way to do it.

This PR also fixes an unrelated small typo in ast_utils.nim.

### DIFF
--- a/src/arraymancer/private/ast_utils.nim
+++ b/src/arraymancer/private/ast_utils.nim
@@ -66,7 +66,7 @@ proc replaceSymsByIdents*(ast: NimNode): NimNode =
       return node
     of nnkLiterals:
       return node
-    of nnkHiddenStdConv: # see `test_fancy_indexing,nim` why needed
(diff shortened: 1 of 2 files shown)


mratsim:
> It would be great if you could somehow refer to the "op" that is used for the reduction or the fold but I did not find a way to do it.

Try `astToStr(op)`

AngelEzquerra:
> > It would be great if you could somehow refer to the "op" that is used for the reduction or the fold but I did not find a way to do it.
> 
> Try `astToStr(op)`

Thank you. I have added the op to the reduce_axis_inline exception. I haven't added it to the fold_axis_inline one because there are 3 ops and I did not want to overcomplicate the exception message.

PR #609, opened 2023-10-21
Doc improvements

AngelEzquerra:
This PR makes a few small improvements to the docs and the README. Each commit is atomic so hopefully they are pretty self explanatory.


mratsim:
Awesome, thanks!

PR #610, opened 2023-10-22
Missing math

AngelEzquerra:
This PR is related to PR #603 which I have closed. Instead of adding "math" versions of the + and - operators, I make those a compilation error which tells the user to use +. and -. instead. In addition to that this PR:
- Adds missing documentation to some of the existing math functions.
- Adds several math functions and operators that exist in std/math and in numpy's math library but were missing from arraymancer.
- Adds a `.*` version of the `tensor * scalar` operator to be consistent with the fact that we already have a `=.*` operator.
- Makes the `./` version of the `tensor / scalar` operator work with integers (in the same way that the `tensor / tensor` version already did).


src/arraymancer/tensor/aggregate.nim
```diff
@@ -337,6 +342,116 @@ proc cumprod*[T](arg: Tensor[T], axis: int = 0): Tensor[T] = # from hugogranstro
     else:
       temp[_] = result.atAxisIndex(axis, i-1) *. tAxis
 
+proc diff*[T](arg: Tensor[T], n=1, axis: int = -1): Tensor[T] =
```
mratsim, line 345: I usually like to reuse Numpy names but it's not very expressive, I think `discrete_diff` would be more expressive.

src/arraymancer/tensor/aggregate.nim
```diff
@@ -337,6 +342,116 @@ proc cumprod*[T](arg: Tensor[T], axis: int = 0): Tensor[T] = # from hugogranstro
...
+      continue
+    var temp = result.atAxisIndex(axis, i-1)
+    when T is bool:
+      temp[_] = tAxis != arg.atAxisIndex(axis, i-1)
+    else:
+      temp[_] = tAxis -. arg.atAxisIndex(axis, i-1)
+  if n > 1:
+    result = diff(result, n=n-1, axis=axis)
+
+proc unwrap*[T: SomeNumber](t: Tensor[T], discont: T = -1, axis = -1, period: T = default(T)): Tensor[T] {.noinit.} =
```
mratsim, line 381: `unwrap` is worse than `diff` because it conflicts with the usual software engineering "unwrap" which is unwraping Option or Result and assuming there was no error.

I'm not sure what's the mathematical name though.

mratsim (commented):
Thank you, unsure about naming of 2 procs

src/arraymancer/tensor/aggregate.nim
AngelEzquerra: I agree that it is not a great name but it is a "traditional" name in the signal processing jargon. Perhaps we could call it "unwrap_period" to make it clear that it unwraps "around" the selected period?

src/arraymancer/tensor/aggregate.nim
AngelEzquerra: I don't find "diff" a confusing name and it is used not just by numpy but also by Matlab and Julia. I can change it if you want but IMHO it is pretty clear.

mratsim: I think `diff_discrete` so that those looking for `diff` get the autocomplete and can see "ah this is it"

src/arraymancer/tensor/aggregate.nim
mratsim: `unwrap_period` sounds good to me.

AngelEzquerra: Change made and PR updated.

src/arraymancer/tensor/aggregate.nim
AngelEzquerra: Change made and PR updated. Also improve the documentation of the function a bit.

src/arraymancer/tensor/aggregate.nim
AngelEzquerra: I rebased this branch on top of the latest master branch and I added a couple more missing math functions that I find useful (isNaN and classify).

src/arraymancer/tensor/aggregate.nim
```diff
@@ -337,6 +342,122 @@ proc cumprod*[T](arg: Tensor[T], axis: int = 0): Tensor[T] = # from hugogranstro
...
+    #
+    # Inputs:
+    #   - t: Input Tensor.
+    #   - discont: Maximum discontinuity between values. Default is `period/2`.
+    #       Values below `period/2` are treated as if they were `period/2`.
+    #       To have an effect different than the default, `discont` must be
+    #       larger than `period/2`.
+    #   - axis: Axis along which the unwrap will be done. Default is the last axis.
+    #   - period: Size of the range over which the input wraps.
+    #             By default, it is ``2*PI``.
```
Vindaar, line 405: Except it is currently `default(T`) :)

src/arraymancer/tensor/aggregate.nim
```diff
@@ -337,6 +342,122 @@ proc cumprod*[T](arg: Tensor[T], axis: int = 0): Tensor[T] = # from hugogranstro
...
+  let axis = if axis == -1:
+    t.shape.len + axis
+  else:
+    axis
+  let td = t.diff_discrete(axis=axis)
+  let period: T = if period == default(T):
+    when T is int:
+      raise newException(ValueError, "unwrap period must be specified for integer types")
+    else:
+      2.0 * PI
```
Vindaar, line 425:
```suggestion
      T(2.0 * PI)
```

possibly.

Vindaar:
Nice work! Those are useful additions indeed. Can't say I spent long enough to fully grasp the `unwrap` logic right now, but I trust your tests and you. :smile: 

I'll merge this after the masked fill is done, ok?

Vindaar:
Thanks again! :rocket:

PR #612, opened 2023-11-01
Add support for doing a masked fill from a tensor

AngelEzquerra:
This is supported by numpy but was not supported by Arraymancer yet.

I'm hoping I've implemented it in a sufficiently performant way. If I can do it in a better way, please let me know.

### DIFF
--- a/docs/tuto.slicing.rst
+++ b/docs/tuto.slicing.rst
@@ -260,3 +260,35 @@ the second to last operation in reverse in the final operation, thus
 copying exactly the thing we copied to the second to last row in
 reverse back to the last row. But because that is where the values in
 the second to last row originated from, nothing "happens".
+
+
+Boolean Masks
+~~~~~~~~~~~~~
+
+In addition to regular slicing, boolean masks can be used to select items
+from a tensor. The mask should have the same shape as the tensor it is used on.
+However, the result of the mask operation will be a flat, 1-D tensor with the
+selected items.
+
+.. code:: nim
+
+    foo = vandermonde.toTensor()
+    echo foo[foo >. 27]
+
+    # Tensor[system.int] of shape "[9]" on backend "Cpu"
+    #     32      81     243      64     256    1024     125     625    3125
+
+Boolean masks can also be used to mutate a tensor. The tensor is mutated in
+place, and thus it maintains its original shape.
+
+.. code:: nim
+
+    foo = vandermonde.toTensor()
+    foo[foo >. 27] = -arange(9)
+
+    # Tensor[system.int] of shape "[5, 5]" on backend "Cpu"
+    # |1      1     1     1     1|
+    # |2      4     8    16     0|
+    # |3      9    27    -1    -2|
+    # |4     16    -3    -4    -5|
+    # |5     25    -6    -7    -8|
--- a/src/arraymancer/tensor/selectors.nim
+++ b/src/arraymancer/tensor/selectors.nim
@@ -19,7 +19,7 @@ import  ./backend/memory_optimization_hints,
         ./private/p_empty_tensors,
         ./accessors, ./accessors_macros_syntax,
         ./data_structure, ./init_cpu,
-        std/sequtils
+        std/[sequtils, locks]
 
 # Indexed axis
 # --------------------------------------------------------------------------------------------
@@ -204,16 +204,32 @@ proc masked_fill*[T](t: var Tensor[T], mask: Tensor[bool], value: Tensor[T]) =
   #     value
   #   else:
   #     x
+
+  # We need to protect against the case in which we run out of values to
+  # fill the tensor with, which we cannot tell in advance without counting
+  # the number of true values in the mask (which would be expensive)
+  var lock: Lock
+  initLock(lock)
+  var too_few_values = false
+
   omp_parallel_blocks(block_offset, block_size, t.size):
     var n = block_offset
-    try:
-      for tElem, maskElem in mzip(t, mask, block_offset, block_size):
-        if maskElem:
-          tElem = value[n]
-          inc n
-    except IndexDefect:
-      raise newException(IndexDefect, "The size of the value tensor (" & $value.size &
-        ") is smaller than the number of true elements in the mask (" & $mask.size & ")")
+    for tElem, maskElem in mzip(t, mask, block_offset, block_size):
+      if maskElem:
+        if n >= value.size:
+          withLock(lock):
+            # The lock protection is technically unnecessary but it is good form
+            too_few_values = true
+          break
+        tElem = value[n]
+        inc n
+  if too_few_values:
+    let error_msg = "masked_fill error: the size of the value tensor (" & $value.size &
+      ") is smaller than the number of true elements in the mask"
+    when not(compileOption("mm", "arc") or compileOption("mm", "orc")):
+      # Other memory management modes crash without showing the exception message
+      echo error_msg
+    raise newException(IndexDefect, error_msg)
 
 proc masked_fill*[T](t: var Tensor[T], mask: openArray, value: Tensor[T]) =
   ## For the index of each element of t.
--- a/src/arraymancer/tensor/selectors.nim
+++ b/src/arraymancer/tensor/selectors.nim
@@ -125,10 +125,9 @@ proc masked_select*[T](t: Tensor[T], mask: openArray): Tensor[T] {.noinit.} =
   t.masked_select mask.toTensor()
 
 proc masked_fill*[T](t: var Tensor[T], mask: Tensor[bool], value: T) =
-  ## For the index of each element of t.
-  ## Fill the elements at ``t[index]`` with the ``value``
-  ## if their corresponding ``mask[index]`` is true.
-  ## If not they are untouched.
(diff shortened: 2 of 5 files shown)


src/arraymancer/tensor/selectors.nim
```diff
@@ -179,6 +178,66 @@ proc masked_fill*[T](t: var Tensor[T], mask: openArray, value: T) =
...
+  omp_parallel_blocks(block_offset, block_size, t.size):
+    var n = block_offset
+    try:
+      for tElem, maskElem in mzip(t, mask, block_offset, block_size):
+        if maskElem:
+          tElem = value[n]
+          inc n
+    except IndexDefect:
+      raise newException(IndexDefect, "The size of the value tensor (" & $value.size &
+        ") is smaller than the number of true elements in the mask (" & $mask.size & ")")
```
mratsim, line 216: Have tried this on Nim 1.6 and Nim 2.0?

This for sure crashes, likely with an incomprehensible error message on Nim 1.6 because exceptions are allocated in a thread-local manner.

In general, OpenMP sections should not allocate.

mratsim (approved):

src/arraymancer/tensor/selectors.nim
AngelEzquerra: You are right. I had tried the "happy path" on both 2.0 and 1.6 but I had only tried the unhappy path in nim 2.0 (in which it works fine, at least with the small tensors that I tried it with). In 1.6 it does cause a crash.
What is the solution? I don't see any error handling on any of the existing OpenMP sections. Even the following (stupid) code in which I create the message and the exception beforehand crashes in 1.6:
```nim
  let error_msg = "The size of the value tensor (" & $value.size &
        ") is smaller than the number of true elements in the mask (" & $mask.size & ")"
  let indexDefectException = newException(IndexDefect, error_msg)

  omp_parallel_blocks(block_offset, block_size, t.size):
    var n = block_offset
    try:
      for tElem, maskElem in mzip(t, mask, block_offset, block_size):
        if maskElem:
          tElem = value[n]
          inc n
    except IndexDefect:
      raise indexDefectException
```
I guess I could have 2 versions of the function, one for 2.0 with error checking and one for 1.6 without it. Is that the way to go?

mratsim: The way to go is to have no exceptions in parallel sections.

And in general, no seq or strings allocation, using only preallocated stuff, otherwise the parallel section needs this: https://github.com/mratsim/laser/blob/e23b5d6/laser/openmp.nim#L88-L110.

```Nim
template attachGC*(): untyped =
  ## If you are allocating reference types, sequences or strings
  ## in a parallel section, you need to attach and detach
  ## a GC for each thread. Those should be thread-local temporaries.
  ##
  ## This attaches the GC.
  ##
  ## Note: this creates too strange error messages
  ## when --threads is not on: https://github.com/nim-lang/Nim/issues/9489
  if(omp_get_thread_num()!=0):
    setupForeignThreadGc()

template detachGC*(): untyped =
  ## If you are allocating reference types, sequences or strings
  ## in a parallel section, you need to attach and detach
  ## a GC for each thread. Those should be thread-local temporaries.
  ##
  ## This detaches the GC.
  ##
  ## Note: this creates too strange error messages
  ## when --threads is not on: https://github.com/nim-lang/Nim/issues/9489
  if(omp_get_thread_num()!=0):
    teardownForeignThreadGc()
```

I'm not sure what's the overhead of attacjing/detaching the GC.

This is the main motivation behind Nim switching to arc/orc GCs.

In the future it will be fine to require Nim 2.0, but I need a better Cuda story (it only works on Nim v1.2 at the moment unless you rebuild a nim.cfg from scratch that replaces --std:gnu++14 by --std:stdc++14.

AngelEzquerra: I have made a change that ensures that when non ARC/ORC memory management the exception error message is shown. I have also added a test for the exception code when using ARC or ORC.

src/arraymancer/tensor/selectors.nim
```diff
@@ -179,6 +178,82 @@ proc masked_fill*[T](t: var Tensor[T], mask: openArray, value: T) =
...
+          withLock(lock):
+            # The lock protection is technically unnecessary but it is good form
+            too_few_values = true
+          break
+        tElem = value[n]
+        inc n
+  if too_few_values:
+    let error_msg = "masked_fill error: the size of the value tensor (" & $value.size &
+      ") is smaller than the number of true elements in the mask"
+    when not(compileOption("mm", "arc") or compileOption("mm", "orc")):
```
Vindaar, line 229: An alterative that could work for older Nim would be:

```suggestion
    when not defined(gcDestructors)
```

but I don't mind to just drop 1.4.

Vindaar:
It looks fine to me now. One could clarify a bit more in the docstrings how different ranks are handled and specify that the mask and fill value tensors must have the same shape. The next thing 'missing' thing would then be a single element fill I guess (which could be convenient e.g. precisely for the type of NaN test case).

Anyway, I think this fine to merge. Please just drop the Nim 1.4 from the CI. I don't see a purpose in continuing to support it officially. Most of it will work regardless.

src/arraymancer/tensor/selectors.nim
AngelEzquerra: I can try this change. In any case I think dropping 1.4 makes sense.

AngelEzquerra:
> It looks fine to me now. One could clarify a bit more in the docstrings how different ranks are handled and specify that the mask and fill value tensors must have the same shape. The next thing 'missing' thing would then be a single element fill I guess (which could be convenient e.g. precisely for the type of NaN test case).
> 
> Anyway, I think this fine to merge. Please just drop the Nim 1.4 from the CI. I don't see a purpose in continuing to support it officially. Most of it will work regardless.

How would I drop Nim 1.4 from the CI? Is that something that I can do myself?

Vindaar:
> > It looks fine to me now. One could clarify a bit more in the docstrings how different ranks are handled and specify that the mask and fill value tensors must have the same shape. The next thing 'missing' thing would then be a single element fill I guess (which could be convenient e.g. precisely for the type of NaN test case).
> > Anyway, I think this fine to merge. Please just drop the Nim 1.4 from the CI. I don't see a purpose in continuing to support it officially. Most of it will work regardless.
> 
> How would I drop Nim 1.4 from the CI? Is that something that I can do myself?

Ah, sorry. I wanted to add that to my response, but I forgot. Yes, just remove it here in this line:

https://github.com/mratsim/Arraymancer/blob/master/.github/workflows/ci.yml#L17

AngelEzquerra:
> It looks fine to me now. One could clarify a bit more in the docstrings how different ranks are handled and specify that the mask and fill value tensors must have the same shape. The next thing 'missing' thing would then be a single element fill I guess (which could be convenient e.g. precisely for the type of NaN test case).

I've pushed a new commit that improves the documentation of all the masked_fill procedures (including the ones that fill with a scalar value).

> Anyway, I think this fine to merge. Please just drop the Nim 1.4 from the CI. I don't see a purpose in continuing to support it officially. Most of it will work regardless.

Nim 1.4 has been dropped on a separate PR.

docs/tuto.slicing.rst
```diff
@@ -260,3 +260,35 @@ the second to last operation in reverse in the final operation, thus
...
+
+    foo = vandermonde.toTensor()
+    foo[foo >. 27] = -arange(9)
+
+    # Tensor[system.int] of shape "[5, 5]" on backend "Cpu"
+    # |1      1     1     1     1|
+    # |2      4     8    16    -1|
+    # |3      9    27    -1    -1|
+    # |4     16    -1    -1    -1|
+    # |5     25    -1    -1    -1|
```
Vindaar, line 294: Why is the output all `-1`? Shouldn't the numbers decrease? Or is it picking the wrong overload and only uses the first element for some reason?

Vindaar: I think the latter might be happening, because this (on current master):
```nim
import arraymancer

var t = arange(25).reshape([5, 5])
t[t >=. 8] = 111
echo t
```
yields
```
Tensor[system.int] of shape "[5, 5]" on backend "Cpu"
|0        1      2      3      4|
|5        6      7    111    111|
|111    111    111    111    111|
|111    111    111    111    111|
|111    111    111    111    111|
```

src/arraymancer/tensor/selectors.nim
```diff
@@ -155,26 +158,111 @@ proc masked_fill*[T](t: var Tensor[T], mask: Tensor[bool], value: T) =
...
+      # Other memory management modes crash without showing the exception message
+      echo error_msg
+    raise newException(IndexDefect, error_msg)
+
+proc masked_fill*[T](t: var Tensor[T], mask: openArray, value: Tensor[T]) =
```
Vindaar, line 243: Is it `openArray` and not `openArray[bool]` for a reason?

docs/tuto.slicing.rst
Vindaar: Ah, I just checked out your branch. It seems to work fine on my end. I assume you maybe just copied the wrong output tensor?

Vindaar:
Really nice work! Could you update the tutorial vandermonde boolean mask example? I'll merge it after. :partying_face:

Issue #487, opened 2021-01-03
Return `var` results when indexing


cmacmackin:
It would be handy to add indexing operators with return type `var Tensor[T]` so that the result can be passed to a function for in-place operations. This may relate to #52.

Vindaar:
Can you clarify what you mean by providing an example?

There are multiple ways to get a mutable value from a `Tensor`:
- using mutable iterators: https://mratsim.github.io/Arraymancer/accessors.html#mitems.i%2CTensor%5BT%5D and friends
- using `[]` on a mutable tensor:
  - `[]` calls `atIndex` here after desugaring: https://github.com/mratsim/Arraymancer/blob/master/src/arraymancer/tensor/private/p_accessors_macros_read.nim#L206
  - which has both a mutable and non mutable overload: https://mratsim.github.io/Arraymancer/p_accessors.html#atIndex%2CTensor%5BT%5D%2Cvarargs%5Bint%5D_2
- using `atContiguousIndex`: https://mratsim.github.io/Arraymancer/accessors.html#atContiguousIndex%2CTensor%5BT%5D%2Cint_2
- using a `RawMutableView`: https://github.com/mratsim/Arraymancer/blob/master/src/arraymancer/laser/tensor/datatypes.nim#L20 by using https://mratsim.github.io/Arraymancer/datatypes.html#unsafe_raw_buf%2CTensor%5BT%3A+KnownSupportsCopyMem%5D%2Cstaticbool_2

And probably more than I'm not aware of right now.

mratsim:
I'm all for it but every time I tried I had nasty bugs that surfaced in the test suite.
Some of which were corrupting the return data iirc.

In particular returning `var seq[T]` is broken and if var seq[T] doesn't work, there is no chance for `var Tensor[T]` to work
https://github.com/mratsim/Arraymancer/pull/420/commits/91850c5e6dd723d29b3b35072f57af4b7988d82d#diff-9827c41e5a724052d30ac0f444f15f465729f1116cdd2614a16f27fd3ad041c4R98

The workaround is to assign the result of indexing to a var first and pass that to the function.

cmacmackin:
@Vindaar When I tried using the `[]` operator in a place where a mutable result was needed I got a compiler-error because apparently it didn't return a mutable result.

@mratsim Is this due to some upstream bug with Nim?

Thanks for the workaround, I've been doing that. Reminds me of using Fortran...

PR #613, opened 2023-11-05
Add at_mut template to enable masked_fill on slices

AngelEzquerra:
This can be useful, for example, when assigning a value into a chain of slice operations which are usually considered immutable even if the original tensor is mutable. For example, the following code:

```nim
var x = arange(20).reshape([4, 3])
x[1..2, _][condition] = 100
```
will fail with a `a slice of an immutable tensor cannot be assigned to` error. Using `at_mut` you can do:
```nim
x.at_mut(1..2, _)[condition] = 100
```
which works fine.

### DIFF
--- a/src/arraymancer/tensor/syntactic_sugar.nim
+++ b/src/arraymancer/tensor/syntactic_sugar.nim
@@ -55,25 +55,3 @@ template at_mut*[T](t: var Tensor[T], args: varargs[untyped]): untyped =
   ##     Singleton dimension are collapsed
   var mt = t[args].squeeze
   mt
-
-template mut*[T](t: Tensor[T]): var Tensor[T] =
-# template mut*[T](t: Tensor[T]): untyped =
-  ## Return a mutable view of a Tensor
-  ## 
-  ## This can be useful, for example, when assigning a value into a chain
-  ## of slice operations which are usually considered immutable even if
-  ## the original tensor is mutable. For example, this lets you do:
-  ##
-  ## .. code:: nim
-  ##   var x = arange(20).reshape([4, 3])
-  ##   # The code `x[1..2, _][condition] = 100` would fail with
-  ##   # a `a slice of an immutable tensor cannot be assigned to` error
-  ##   # Using `mut` allows assignment to the slice
-  ##   x[1..2, _].mut[condition] = 100
-  ##
-  ## Input:
-  ##   - a Tensor
-  ## Returns:
-  ##   - a mutable value or view of the Tensor corresponding to the slice
-  var mt = t
-  mt
--- a/src/arraymancer/tensor/syntactic_sugar.nim
+++ b/src/arraymancer/tensor/syntactic_sugar.nim
@@ -29,3 +29,51 @@ template at*[T](t: Tensor[T], args: varargs[untyped]): untyped =
   ## Usage:
   ##   See the ``[]`` macro
   t[args].squeeze
+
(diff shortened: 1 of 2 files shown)


src/arraymancer/tensor/syntactic_sugar.nim
```diff
@@ -29,3 +29,51 @@ template at*[T](t: Tensor[T], args: varargs[untyped]): untyped =
...
+  ##   - and:
+  ##     - specific coordinates (``varargs[int]``)
+  ##     - or a slice (cf. tutorial)
+  ## Returns:
+  ##   - a mutable value or view of the Tensor corresponding to the slice
+  ##     Singleton dimension are collapsed
+  var mt = t[args].squeeze
+  mt
+
+template mut*[T](t: Tensor[T]): var Tensor[T] =
```
mratsim, line 59: This is breaking assumptions and should be named unsafe_mut, but I think `at_mut` would cover all use-cases no?

AngelEzquerra: I think you are right on both accounts (i.e. it is unsafe and we might not need it).
I was looking for the best way to indicate that the slice can be mutated. `at_mut` probably covers the cases I've uncovered so far so I guess I could remove `mut` for now and we could reconsider it in case we find a case that is not covered by `at_mut`.

That being said, I find that the syntax using `mut` looks better as it does not force you to switch from the `[]=` operator to the `at_mut` function call syntax. If only there was a way to use `mut` but have it fail to compile when the target Tensor were immutable... That is, if we could support:
```nim
var x = arange(12).reshape([4, 3])
x[1..2, _].mut[condition] = 100
```
while this failed to compile:
```nim
let x = arange(12).reshape([4, 3])
x[1..2, _].mut[condition] = 100
```
Then `mut` (which would no longer be unsafe) would be the superior solution IMHO.
If you have any ideas for that let me know, otherwise we can just keep `at_mut`.

mratsim: There are 2 solutions to this.

The first one and the most ergonomic one is to ensure that when slicing `var Tensor` you return a `var Tensor`.

I started and abandoned `var slice` 5 years ago: https://github.com/mratsim/Arraymancer/commit/9b5539079dbf34535a5372f80dad980a0059d990
and iirc try to resurrect them with Nim ``--experimental:views`` support but I was still kind of stuck by the type system, see git blame: https://github.com/mratsim/Arraymancer/blame/f809e7f2806020f4c49ca45f000e478fe53fd2ba/src/arraymancer/tensor/private/p_accessors_macros_write.nim#L262

The second one would be quite a burden. It is actually used internally to distinguish between mutable and immutable memory when using pointers:

You use distinct types and you implement only mutable procs on the "Mut" variant:
https://github.com/mratsim/Arraymancer/blob/f809e7f2806020f4c49ca45f000e478fe53fd2ba/src/arraymancer/laser/tensor/datatypes.nim#L24-L25

This works well for internal pointers because pointer only support indexing reads and indexing writes. For tensors, there are too many functions and I don't thing there is a way to make that "TensorViewImmutable" disappear and becoming a simple "Tensor" when assigned to `let`, i.e. it would only exist in flight.

It should be possible via a macro, say we call it `chain` or `chainMut` and internally we assign to `var` but code becomes quite unintuitive.

mratsim: See also #487

AngelEzquerra:
I've removed the unsafe mut template.

PR #617, opened 2024-01-05
Convolution

AngelEzquerra:
Add a convolve function with similar API and behavior as numpy's convolve.
I added openmp support. This is the first time I use it so hopefully I did it right!

### DIFF
--- a/src/arraymancer/tensor/math_functions.nim
+++ b/src/arraymancer/tensor/math_functions.nim
@@ -113,3 +113,45 @@ proc square*[T](x: T): T {.inline.} =
   x*x
 
 makeUniversal(square)
+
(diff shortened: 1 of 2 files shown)


mratsim (approved):
Thanks.

Does signal processing also use cross-entropy?

AngelEzquerra:
> Thanks.
> 
> Does signal processing also use cross-entropy?

Not in my experience.

I might make another PR later to add a correlation function which numpy supports as well. If we add an FFT too at some point (which I think we should since it’s such a foundational DSP algorithm) we could use it to calculate the convolution more efficiently for larger tensor sizes.

mratsim:
Ah yes I meant cross-correlation not cross-entropy

PR #619, opened 2024-01-11
Missing numpy functions

AngelEzquerra:
This PR adds a number of missing numpy functions:
- tril, triu: These were already implemented but not exported nor tested
- tri
- geomspace
- sinc
- meshgrid
- all / any
- flatten / roll

I'd like to add more but I think this is a good first set.


mratsim (approved):
Thanks, excellent

AngelEzquerra:
I rebased this branch on top of the latest master branch.

AngelEzquerra:
Added missing `flatten` and `roll` functions.

Vindaar:
Also great work on this! :partying_face: 

After merging the other two (and having dropped 1.4 support), CI should pass for this one too. Can never remember whether you'll need to rebase onto the then-master without 1.4 though.

PR #621, opened 2024-02-12
Remove support for nim 1.4

AngelEzquerra:
A few incoming PR's become simpler if we drop support for nim 1.4, which according to the most recent Nim Community Survey is used by less than 1.5% of nim's users.


AngelEzquerra:
Thanks to @Vindaar for telling me how to do this change.

Vindaar:
Very good and thanks!

PR #622, opened 2024-02-16
More missing numpy

AngelEzquerra:
This adds a few more missing numpy functions: append, moveaxis, vander (similar to vandermonde by with reversed column order by default) and several scalar-tensor comparison operators (until now we only supported tensor-tensor and tensor-scalar versions of those operators, but not scalar-tensor).

### DIFF
--- a/src/arraymancer/linear_algebra/special_matrices.nim
+++ b/src/arraymancer/linear_algebra/special_matrices.nim
@@ -34,21 +34,90 @@ proc hilbert*(n: int, T: typedesc[SomeFloat]): Tensor[T] =
   # TODO: scipy has a very fast hilbert matrix generation
   #       via Hankel Matrix and fancy 2D indexing
 
-proc vandermonde*[T](x: Tensor[T], order: int): Tensor[float] =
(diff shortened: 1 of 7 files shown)


src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -34,21 +34,90 @@ proc hilbert*(n: int, T: typedesc[SomeFloat]): Tensor[T] =
...
+  ## A "generalized" Vandermonde matrix consists of the input `x` split into
+  ## multiple rows where each row contains the powers of `x_i` elevated to each
+  ## of the elements of the `orders` tensor.
   ##
   ## `V_ij = x_i ^ order_j`
   ##
-  ## where `order_j` runs from 0 to `order`.
+  ## Inputs:
+  ##   - x: The input tensor `x` (which must be a rank-1 tensor)
+  ##   - orders: The "exponentes" tensor (which must also be a rank-1 tensor)
```
Vindaar, line 48:
```suggestion
  ##   - orders: The "exponents" tensor (which must also be a rank-1 tensor)
```

Vindaar:
Thanks again! :partying_face:

PR #623, opened 2024-02-17
Add Tensor.len as an alias for Tensor.size returning an int

AngelEzquerra:
This will make writing code that works both for tensors and openArrays much easier.

This also adds a docstring for `size`.


Vindaar:
And yet again, thanks! :grin:

PR #624, opened 2024-02-17
Make masked_fill work with openArray values

AngelEzquerra:
This makes it possible to do a masked_fill with a seq or array as the values without having to convert them into a tensor first (which could be expensive for large tensors).
This is something that numpy also supports. For example:
```python
import numpy as np
a = np.array([[-1, -2, 1], [1, 2, 0], [1, -1, 1]])
a[a < 0] = [-100, -200, -300]
```
Note how we didn't need to convert [-100, -200, -300] to an array before assigning it to `a`.

### DIFF
--- a/src/arraymancer/tensor/selectors.nim
+++ b/src/arraymancer/tensor/selectors.nim
@@ -180,25 +180,17 @@ proc masked_fill*[T](t: var Tensor[T], mask: openArray, value: T) =
     return
   t.masked_fill(mask.toTensor(), value)
 
-proc masked_fill*[T](t: var Tensor[T], mask: Tensor[bool], value: Tensor[T]) =
-  ## For each element ``t[index]`` of the input tensor ``t`` with index ``index``,
-  ## check if ``mask[index]`` is true. If so fill it with the _next_
-  ## element from the ``value`` tensor. Otherwise leave it untouched.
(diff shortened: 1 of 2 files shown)


src/arraymancer/tensor/selectors.nim
```diff
@@ -255,17 +247,48 @@ proc masked_fill*[T](t: var Tensor[T], mask: openArray, value: Tensor[T]) =
...
+  ##
+  ## Example:
+  ##
+  ##   t.masked_fill(t > 0, [3, 4, -1])
+  ##
+  ## In this version of this procedure the boolean mask is a ``Tensor[bool]``
+  ## with the same size as the input tensor ``t``.
+  masked_fill_impl(t, mask, value)
 
+proc masked_fill*[T](t: var Tensor[T], mask: openArray, value: Tensor[T]) =
```
Vindaar, line 279: Does explicit `openArray[bool]` break?

AngelEzquerra: Yes, openArray[bool] does not work because the mask can be, for example, a seq[seq[bool]]. Note the aim not changing this, it was already like that in the orirginal code.

Vindaar: Yup, but I forgot to ask I think. Good to know and that makes sense! :)

PR #625, opened 2024-02-19
Relax some shape checks for rank-1 tensors

AngelEzquerra:
Consider that rank-1 tensors are "shapeless" and can fit both into [1, N] and [N, 1] tensors, for example.

This greatly reduces the need to use .reshape after converting a seq or array into a tensor in assignments, and reduces errors that happen at runtime for code that is conceptually correct. For example, with this change this will work:
```
var t = eye(3, 3)
t[_,0] = [1, 2, 3].toTensor
```

### DIFF
--- a/tests/tensor/test_fancy_indexing.nim
+++ b/tests/tensor/test_fancy_indexing.nim
@@ -122,6 +122,31 @@ proc main() =
 
         check: y == exp
 
+    test "Masked assign tensor or openArray via fancy indexing":
+      block: # y[y > 50] = np.array([-100, -200])
+        var y = x.clone()
+        # Assing a tensor
+        y[y >. 50] = [-100, -200].toTensor()
+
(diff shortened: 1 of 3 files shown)


tests/tensor/test_fancy_indexing.nim
```diff
@@ -175,6 +200,19 @@ proc main() =
 
         check: y == exp
 
+      block:
+        # Assigning a rank-1 tensor into an axis of the same size is supported
+        # Note that this is not supported by numpy
```
Vindaar, line 205: Does numpy have a good reason not to support this maybe?

Vindaar:
Having digested it now, I like it. As it is a breaking change, I made the hardcoded `true` value based on a CT variable, `RelaxedRankOne`, which can be disabled by compiling with `-d:RelaxedRankOne=false` to recover the old behavior.

PR #628, opened 2024-02-23
Add int2bit functions

AngelEzquerra:
These let you convert integers or tensors of intengers into tensors of bools representing them. This is similar to Matlab's `int2bit` (except that Matlab's version fills the bit tensors column-wise, while arraymancer's version fills them row-wise). It is also similar to (but more flexible than) numpy's `unpackbits`.


Vindaar:
If we ever add more signal processing related things to arraymancer directly, we might want to add a `signals` (or similar) directory and have it there? For now `special_matrices` is fine, I suppose.

PR #629, opened 2024-02-28
Support complex64 tensor - real scalar operations (like numpy does)

AngelEzquerra:
It is always possible to operate on a Complex64 tensor with a real scalar without loss of precision, so we allow it without explicit type conversion.

This is supported by numpy, and makes complex tensor arithmetic more convenient to use (by avoiding having to explicitly make many obvious conversions to complex).

As an example, assuming `t` is a Tensor[Complex64], this let's you write expressions such as:
```
(complex(1.0 / sqrt(2.0)) *. t -. complex(1.0)) /. complex(t.size.float)
```

as follows:
```
((1.0 / sqrt(2.0)) *. t -. 1.0) /. t.size
```

which is much more easier to understand while being equally safe.


Vindaar:
Nice, I agree that's useful. I barely use complex numbers in my code, so I never much stumbled over it.

PR #630, opened 2024-02-28
Complex enhancements

AngelEzquerra:
A couple of improvements to the complex() function:

- It is now possible to call it without the second (imaginary) tensor. This let's you use similar code whether you are creating a complex tensor by combining 2 tensors or by taking just a single, real tensor. It also has the advantage over `asType` that it automatically chooses the right Complex type based on the input, leading to more generic code.
- It is now possible to use complex() to directly convert integer tensors into complex tensors. This avoids unnecessary copies and also leads to more uniform code when dealing with complex tensors.

### DIFF
--- a/src/arraymancer/tensor/complex.nim
+++ b/src/arraymancer/tensor/complex.nim
@@ -8,18 +8,28 @@ import
   ./accessors,
   ./higher_order_applymap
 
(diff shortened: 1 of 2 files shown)


Vindaar:
Thank you!

PR #631, opened 2024-03-02
improve docstrings of `makeUniversal` templates, allow extensions

Vindaar:
Each `makeUniversal` call accepts a `docSuffix` argument, which can be used to provide an additional piece of text to be added to that calls doc.

@AngelEzquerra: Feel free to adjust any of the `makeUniversal` calls in `ufunc.nim` and give some extra information. It will just be added as a paragraph to the autogenerated text.

Note: The previous comment 

```
# Note, the makeUniversal/Local documentation gets duplicated in docs at each template call
# And shouldn't use ##
```
in the code seems to be outdated. This does not seem to happen anymore. Therefore I've made the doc string of the two templates regular `##` doc comments.


AngelEzquerra (approved):
I just tested it and it works great! Thank you!

PR #636, opened 2024-03-17
Add a few more items to .gitignore

AngelEzquerra:
Add a few more items (mainly windows related items) to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,12 @@ docs/build/
 .cache
 .DS_Store
 .vscode
+.idea/
+htmldocs/
+*.exe
+*.dll
+*.orig
+*.bak
+*.bck
+* - Copy.*
+* - Copy (?).*


.gitignore
```diff
@@ -8,3 +8,12 @@ docs/build/
...
 .vscode
+.idea/
+htmldocs/
+*.exe
+*.dll
+*.orig
+*.bak
+*.bck
+* - Copy.*
+* - Copy (?).*
```
Vindaar, line 19: Is that valid syntax? Wouldn't it be something like `*\ -\ Copy.*`?

AngelEzquerra: I believe it is (and it does work on Windows). According to `gitignore`'s documentation:

> Trailing spaces are ignored unless they are quoted with backslash ("\").

As far as I can tell that is the only place where you need to add backslashes.

PR #640, opened 2024-03-21
Fix issue 639

AngelEzquerra:
To fix this issue I made 2 sets of changes:

1. Add a check for rank == 0 to the DynamicStackArray's `product` function (which fixes the actual issue).
2. Make it (a bit) harder to accidentally create a rank-0 tensor (by making newTensor and newTensorUninit create an empty rank-1 tensor by default).

It is still possible (and sometimes useful) to create a rank-0 tensor, but users will need to do so explicitly.


Vindaar:
Great! :partying_face:

PR #645, opened 2024-04-03
Add a `unique` function to algorithms.nim

AngelEzquerra:
This function is similar to but not as fully featured as `numpy.unique`. It is missing a way to count the number or returning the indexes of the unique elements. It can however optionally sort the output.


Vindaar:
Nice!

PR #646, opened 2024-04-13
Add `reshape_infer` procedure

AngelEzquerra:
Unlike numpy, `reshape` does not support having dimensions with value -1 to infer their value. To do so a new `reshape_infer` is added.

This is added as a separate procedure to avoid the (small) cost this adds on top of the usual reshape (which could be called relatively frequently).


Vindaar:
Cool, that's useful indeed!

PR #647, opened 2024-04-14
Missing special matrices

AngelEzquerra:
Add 3 types of special matrices that were missing from numpy / scipy. All of them are "diagonal related":

- Circulant: A square matrix in which each column (or row) is a cyclic shift of the previous column (or row)
- Toeplitz: A matrix that has constant diagonals, with c as its first column and r as its first row (note that r[0] is ignored but should be the same as c[^1]).
- Hankel: A matrix that has constant anti-diagonals, with c as its first column and r as its last row (note that r[0] is ignored but should be the same as c[^1]).

### DIFF
--- a/src/arraymancer/linear_algebra/special_matrices.nim
+++ b/src/arraymancer/linear_algebra/special_matrices.nim
@@ -304,6 +304,7 @@ proc circulant*[T](t: Tensor[T], axis = -1, step = 1): Tensor[T] {.noInit.} =
   ##         each time. Defaults to 1.
   ## Result:
   ## - The constructed circulant matrix
+  ##
   ## Example:
   ## ```nim
   ## echo circulant([1, 3, 6])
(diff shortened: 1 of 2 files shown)


src/arraymancer/linear_algebra/special_matrices.nim
```diff
@@ -286,6 +286,174 @@ proc tri*[T](shape_ax1, shape_ax0: int, k: static int = 0, upper: static bool =
...
+  ## - The constructed Toeplitz matrix
+  ##
+  ## Notes:
+  ## - There is a version of this procedure that takes a single argument `c`,
+  ##   in which case `r` is set to `c.conjugate`.
+  ## - Use the `hankel` procedure to generate a Hankel matrix instead.
+  ##
+  ## Example:
+  ## ```nim
+  ## echo toeplitz([1, 3, 6], [9, 10, 11, 12])
```
mratsim, line 345: Is that normal that 9 doesn't appear in the resulting example?

mratsim (approved):
LGTM except for an example that might be wrong?

src/arraymancer/linear_algebra/special_matrices.nim
AngelEzquerra: Yes, actually it is normal that 9 does not appear. The example tries to highlight that r[0] is ignored and that while it should be the same as c[0], that is not enforced. If you think that makes the example confusing I could change it.

Vindaar:
Looks good! Never heard of these before either. :)

PR #648, opened 2024-04-22
Add `tile` and `repeat_values` procedures

AngelEzquerra:
Add the following new procedures:

- `repeat_values`: Procedures that let you repeat the values of a Tensor multiple times. This functionality exists both in numpy (`repeat`) and in Matlab (`repelem`).
- `tile`: Procedure that lets you construct a new tensor by repeating the input tensor a number of times on one or more axes. This functionality exists both in numpy (`tile`) and in Matlab (`repmat`).

I didn't follow numpy's naming convention for `repeat_values` to avoid a confusing with nim's `sequtils.repeat`, which does not repeat individual values but the whole sequence (like `tile` does).

I measured the performance of these functions and it is comparable to numpy's. `repeat_values` is faster than numpy in --d:danger mode and a bit slower in --d:release mode. `tile` is slower in both cases (but the difference is not too large). `tile`'s implementation could be improved (to avoid intermediate tensor allocations) in the future.

### DIFF
--- a/src/arraymancer/tensor/shapeshifting.nim
+++ b/src/arraymancer/tensor/shapeshifting.nim
@@ -687,3 +687,67 @@ proc repeat_values*[T](t: Tensor[T], reps: Tensor[int]): Tensor[T] {.noinit, inl
   ## version.
   ## ```
   t.repeat_values(reps.toSeq1D)
+
+proc tile*[T](t: Tensor[T], reps: varargs[int]): Tensor[T] =
+  ## Construct a new tensor by repeating the input tensor a number of times on one or more axes
+  ##
+  ## Inputs:
+  ##   - t: The tensor to repeat
+  ##   - reps: One or more integers indicating the number of times to repeat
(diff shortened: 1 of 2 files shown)


src/arraymancer/tensor/shapeshifting.nim
```diff
@@ -571,3 +571,183 @@ proc roll*[T](t: Tensor[T], shift: int, axis: Natural): Tensor[T] {.noinit.} =
       let result_idx = floorMod(n + shift, t.shape[axis])
       rolled_slices[result_idx] = t_slice
     result = concat(rolled_slices, axis)
+
+proc repeat_values*[T](t: Tensor[T], reps: int, axis = -1): Tensor[T] {.noinit.} =
+  ## Create a new tensor with each value repeated (the same amount of)`reps` times
```
Vindaar, line 576:
```suggestion
  ## Create a new tensor with each value repeated (the same amount of) `reps` times
```

PR #649, opened 2024-04-25
More missing numpy and matlab

AngelEzquerra:
This PR adds multiple functions that are available both in numpy and Matlab:

- `union`: returns the unique, unsorted Tensor of values that are found in either of the two input Tensors.
- `intersection`: returns the "intersection" of 2 Tensors as an unsorted rank-1 Tensor.
- `setDiff`: returns the (symmetric or non symmetric) "difference" between 2 Tensors as an unsorted rank-1 Tensor.
- `toHashSet`: convert a Tensor into HashSet.
- `contains` (and thus `in` and `notin`).
- ~`sub2ind` and `ind2sub`: convert "subscript" sequences into linear tensor indexes and vice-versa.~
- `almostEqual`: missing from nim's std/math.

### DIFF
--- a/src/arraymancer/laser/tensor/initialization.nim
+++ b/src/arraymancer/laser/tensor/initialization.nim
@@ -15,9 +15,10 @@ import
 # Standard library
 import std / [typetraits, sequtils, sets]
 
-# The folling export is needed to avoid an compilation error in
+# The following export is needed to avoid a compilation error in
 # algorithms.nim/intersection() when running the test_algorithms test:
 # `Error: type mismatch - Expression: items(s1)`
+# (Alternative: could use `bind sets.items` in `intersection` and `setDiff`)
 export sets
 
 # Third-party
--- a/src/arraymancer/laser/tensor/initialization.nim
+++ b/src/arraymancer/laser/tensor/initialization.nim
@@ -13,7 +13,7 @@ import
   ../private/nested_containers,
   ./datatypes
 # Standard library
-import std / [typetraits, sequtils]
+import std / [typetraits, sequtils, sets]
 # Third-party
 import nimblas
 
@@ -210,11 +210,11 @@ proc newTensor*[T](shape: Metadata): Tensor[T] =
 
 proc toTensor[T](a: openArray[T], shape: Metadata): Tensor[T] =
   ## Convert an openArray to a Tensor
+  ##
   ## Input:
   ##      - An array or a seq, must be flattened. Called by `toTensor` below.
   ## Result:
   ##      - A Tensor of the same shape
-  ##
   var data = @a
   if unlikely(shape.product != data.len):
     raise newException(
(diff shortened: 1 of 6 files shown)


src/arraymancer/laser/tensor/initialization.nim
```diff
@@ -250,6 +257,18 @@ proc toTensor*[T](a: openArray[T]): auto =
...
+  ## Convert a HashSet or an OrderedSet into a Tensor
+  ##
+  ## Input:
+  ##      - An HashSet or an OrderedSet
+  ## Result:
+  ##      - A Tensor of the same shape
+  var shape = MetaData()
+  shape.add(a.len)
+  let data = toSeq(a)
+  result = toTensor(data, shape)
```
Vindaar, line 270: I suppose we could also do:
```nim
result = newTensorUninit[T](a.card)
for i, x in a:
  result[i] = x
```
to avoid the double copy (set -> seq -> tensor)

AngelEzquerra: Makes sense. I'll make that change

AngelEzquerra:
> `sub2ind` is essentially `getIndex` https://github.com/mratsim/Arraymancer/blob/master/src/arraymancer/tensor/private/p_accessors.nim#L60, no? (just with `seq` input)
> 
> Maybe we can reuse the internal logic? In general `sub2ind` and `ind2sub` seem very Matlab focused concepts / names? I can see how it can be useful sometimes though. In Julia apparently they have since deprecated these for a `CartesianIndex` / `LinearIndex` object helper.

You are right. Let me remove those from this PR. We can discuss offline and I can create another separate PR later.

AngelEzquerra:
> > `sub2ind` is essentially `getIndex` https://github.com/mratsim/Arraymancer/blob/master/src/arraymancer/tensor/private/p_accessors.nim#L60, no? (just with `seq` input)
> > Maybe we can reuse the internal logic? In general `sub2ind` and `ind2sub` seem very Matlab focused concepts / names? I can see how it can be useful sometimes though. In Julia apparently they have since deprecated these for a `CartesianIndex` / `LinearIndex` object helper.
> 
> You are right. Let me remove those from this PR. We can discuss offline and I can create another separate PR later.

I've updated the PR as discussed here.

src/arraymancer/laser/tensor/initialization.nim
AngelEzquerra: Interestingly a performance measurement using `timeit` showed that the original version was faster on Windows. We will leave this as is for now while we investigate what is going on (and if we find out why we'll update this in a separate PR).

AngelEzquerra:
I just pushed a fix to the original (non gemm) version which solves a problem with convolving a slice (was using `unsafe_raw_data` where I should have used `unsafe_raw_offset`).

src/arraymancer/laser/tensor/initialization.nim
```diff
@@ -13,7 +13,13 @@ import
   ../private/nested_containers,
   ./datatypes
 # Standard library
-import std / [typetraits, sequtils]
+import std / [typetraits, sequtils, sets]
+
+# The folling export is needed to avoid an compilation error in
+# algorithms.nim/intersection() when running the test_algorithms test:
+# `Error: type mismatch - Expression: items(s1)`
+export sets
```
Vindaar, line 21: Note that we could also get away with adding a `bind sets.items` inside of `intersection` and `setDiff`. However, given that we now depend on `sets`, I don't see an issue exporting it too.

src/arraymancer/laser/tensor/initialization.nim
Vindaar: Yeah, let's investigate it further after this PR.

src/arraymancer/laser/tensor/initialization.nim
```diff
@@ -13,7 +13,13 @@ import
   ../private/nested_containers,
   ./datatypes
 # Standard library
-import std / [typetraits, sequtils]
+import std / [typetraits, sequtils, sets]
+
+# The folling export is needed to avoid an compilation error in
+# algorithms.nim/intersection() when running the test_algorithms test:
+# `Error: type mismatch - Expression: items(s1)`
```
Vindaar, line 20:
```suggestion
# The following export is needed to avoid a compilation error in
# algorithms.nim/intersection() when running the test_algorithms test:
# `Error: type mismatch - Expression: items(s1)`
# (Alternative: could use `bind sets.items` in `intersection` and `setDiff`)
```

Vindaar:
Merging, we can look into the `toTensor` conversion from a `HashSet` performance later.

PR #650, opened 2024-05-01
Improve the performance of convolve (and correlate)

AngelEzquerra:
The exising implementation of `convolve` (which is also used for `correlate`) was pretty slow. The slowness was due to the fact that we were using regular tensor element access in the inner loop of the convolution, which was very expensive.

The solution was to ensure that the input tensors were contiguous (by cloning them if necessary) and then using `unsafe_raw_buf` for all the tensor element accesses, which is safe because we know that the indexes are all within the tensor boundaries.

On my system this change makes a large convolution go from ~1.5 seconds in release mode and ~0.25 seconds in danger mode down to ~65 usecs in both modes (i.e. a x23 reduction in release mode and a x3.8 reduction in danger mode)!

### DIFF
--- a/src/arraymancer/tensor/math_functions.nim
+++ b/src/arraymancer/tensor/math_functions.nim
@@ -328,12 +328,25 @@ proc convolveImpl[T: SomeNumber | Complex32 | Complex64](
   # Initialize the result tensor
   result = zeros[T](len_result)
 
+  # Ensure that the input tensors are contiguous so that they can be accessed
+  # efficiently using `unsafe_raw_buf` in the inner loop
+  let f = if f.isContiguous(): f else: f.clone()
+  let g = if g.isContiguous(): g else: g.clone()
+
   # And perform the convolution
   omp_parallel_blocks(block_offset, block_size, len_result):
     for n in block_offset ..< block_offset + block_size:
       let shift = n + offset
       for m in max(0, shift - g.size + 1) .. min(f.size - 1, shift):
-        result[n] += f[m] * g[shift - m]
+        # We want to do the following operation:
+        # result[n] += f[m] * g[shift - m]
+        # In order to do it efficently, we want to avoid all the overhead of
+        # using regular `[]` access. Since we know that we are working with
+        # continuous, rank-1 tensors, that `n`, `m` and `shift-m` are within
+        # the boundaries of `result`, `f` and `g` (respectively) and since
+        # `T is KnownSupportsCopyMem`, it is safe (and way more efficient) to
+        # use `unsafe_raw_buf` to access the actual tensor elements here:
+        result.unsafe_raw_offset[n] += f.unsafe_raw_offset[m] * g.unsafe_raw_offset[shift - m]
 
 proc convolve*[T: SomeNumber | Complex32 | Complex64](
     t1, t2: Tensor[T],
--- a/src/arraymancer/tensor/math_functions.nim
+++ b/src/arraymancer/tensor/math_functions.nim
@@ -402,10 +402,6 @@ proc correlateImpl[T](f, g: Tensor[T],
   # otherwise the `gemm` call below doesn't do the right thing!
   result = newTensorUninit[T](1, result_len)
 
-  # # Let's make sure both inputs are contiguous
-  # let f = f.asContiguous()
-  # let g = g.asContiguous()
-
   # Create the "shifted column input matrix" that will be used to calculate the
   # convolution through a matrix multiplication with the kernel
   var input_shifts = newTensorUninit[T](g.len, result_len)
--- a/src/arraymancer/tensor/math_functions.nim
+++ b/src/arraymancer/tensor/math_functions.nim
@@ -16,7 +16,8 @@ import  ./data_structure,
         ./backend/openmp,
         ./init_cpu,
         ./higher_order_applymap,
-        ./ufunc
+        ./ufunc,
+        ./operators_blas_l2l3
 import std / math
 import complex except Complex64, Complex32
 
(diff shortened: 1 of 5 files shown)


mratsim:
Can you check the speed of cross-correlation vs im2colgemm_conv2d: https://github.com/mratsim/Arraymancer/blob/14486980c7bc46bd688467d80216d9a55844e3f3/src/arraymancer/nn_primitives/fallback/conv.nim#L81-L106

You can put a batch size of 1 and number of color channel to 1, result should be the same then.

I think it corresponds to "valid" for convolvemode

AngelEzquerra:
> Can you check the speed of cross-correlation vs im2colgemm_conv2d: https://github.com/mratsim/Arraymancer/blob/14486980c7bc46bd688467d80216d9a55844e3f3/src/arraymancer/nn_primitives/fallback/conv.nim#L81-L106
> 
> 
> 
> You can put a batch size of 1 and number of color channel to 1, result should be the same then. Bias can be an empty Tensor.
> 
> 
> 
> I think it corresponds to "valid" for convolvemode

I have this a try and as you said `im2colgemm_conv2d` with the right arguments is equivalent to `correlate` called with `valid` mode. I made some measurements and the results are a little surprising. When working with some relatively large integer tensors (`arange(50000)` as the input tensor and `arange(200)` as the "filter" tensor) the performance of the version of convolve in this PR and `im2colgemm_conv2d` is basically the same (around 70 ms). However the same test with the same inputs converted to float gave very different results. This PR's convolve's performance worsened to ~100 ms, while `im2colgemm_conv2d` improved to just below 30 ms.

For reference scipy's `upfirdn`'s performance is still much better _in many cases_ but seems a little inconsistent (depending on the order of the inputs it can be worse). I'll need to do more tests to fully confirm this point and provide actual comparison measurements.

In any case it seems worth looking into using `im2colgemm_conv2d`'s method. However, it is not clear to me if it'd be easy to extend its result to "full mode". Any thoughts on that @mratsim? And any ideas about how Scion might be even faster?

AngelEzquerra:
> > Can you check the speed of cross-correlation vs im2colgemm_conv2d: https://github.com/mratsim/Arraymancer/blob/14486980c7bc46bd688467d80216d9a55844e3f3/src/arraymancer/nn_primitives/fallback/conv.nim#L81-L106
> > 
> > You can put a batch size of 1 and number of color channel to 1, result should be the same then. Bias can be an empty Tensor.
> > I think it corresponds to "valid" for convolvemode
> 
> I have this a try and as you said `im2colgemm_conv2d` with the right arguments is equivalent to `correlate` called with `valid` mode. I made some measurements and the results are a little surprising. When working with some relatively large integer tensors (`arange(50000)` as the input tensor and `arange(200)` as the "filter" tensor) the performance of the version of convolve in this PR and `im2colgemm_conv2d` is basically the same (around 70 ms). However the same test with the same inputs converted to float gave very different results. This PR's convolve's performance worsened to ~100 ms, while `im2colgemm_conv2d` improved to just below 30 ms.
> 
> For reference scipy's `upfirdn`'s performance is still much better _in many cases_ but seems a little inconsistent (depending on the order of the inputs it can be worse). I'll need to do more tests to fully confirm this point and provide actual comparison measurements.
> 
> In any case it seems worth looking into using `im2colgemm_conv2d`'s method. However, it is not clear to me if it'd be easy to extend its result to "full mode". Any thoughts on that @mratsim? And any ideas about how scipy might be so much faster?

I saw that is is possible to replicate `convolution`'s `full` and `valid` modes using the padding argument of `im2colgemm_conv2d`. It should also be possible to use the `stride` argument to efficiently replicate the `down` argument of scipy's `upfirdn` function.

I still don't know why `upfirdn` is sometimes much faster than `im2colgemm_conv2d` though.

Vindaar:
> When working with some relatively large integer tensors (`arange(50000)` as the input tensor and `arange(200)` as the "filter" tensor) the performance of the version of convolve in this PR and `im2colgemm_conv2d` is basically the same (around 70 ms). However the same test with the same inputs converted to float gave very different results. This PR's convolve's performance worsened to ~100 ms, while `im2colgemm_conv2d` improved to just below 30 ms.

Possibly related to the fact that it uses `gemm` and for integer types it uses Mamy's laser backend, while for float it falls back to BLAS? iirc the laser `gemm` implementation only exists for integers. 

Regarding the scipy implementation: you are compiling with OpenMP explicitly (`-d:openmp`), no? Easy to forget and I assume scipy will do so.

AngelEzquerra:
> > When working with some relatively large integer tensors (`arange(50000)` as the input tensor and `arange(200)` as the "filter" tensor) the performance of the version of convolve in this PR and `im2colgemm_conv2d` is basically the same (around 70 ms). However the same test with the same inputs converted to float gave very different results. This PR's convolve's performance worsened to ~100 ms, while `im2colgemm_conv2d` improved to just below 30 ms.
> 
> Possibly related to the fact that it uses `gemm` and for integer types it uses Mamy's laser backend, while for float it falls back to BLAS? iirc the laser `gemm` implementation only exists for integers.
> 
> Regarding the scipy implementation: you are compiling with OpenMP explicitly (`-d:openmp`), no? Easy to forget and I assume scipy will do so.

That seems a plausible explanation, thanks. Actually, I wasn't using that option. Unfortunatelly when I compile with `-d:openmp` (on windows) I get the following error:

> initialization.nim: In function 'copyFromRaw__upfirdn_u348':
> initialization.nim:287:63: error: invalid branch to/from OpenMP structured block
> initialization.nim:287:213: error: invalid branch to/from OpenMP structured block
> initialization.nim:289:63: error: invalid branch to/from OpenMP structured block
> compilation terminated due to -fmax-errors=3.


> edit: The code for `upfirdn` doesn't look very complicated: https://github.com/scipy/scipy/blob/main/scipy/signal/_upfirdn.py

The key is that to implement it efficiently you must do the downsampling as part of the convolution. You cannot do it _after_ the convolution step because it's much less efficient (you'd calculate "down" times more samples than needed). I was originally planning to add an `upfirdn` function to impulse, but because of this problem I think it makes more sense to add it to arraymancer. We can add the filter window creation (e.g. kaiser, etc) to impulse, and use them to add a `resample` function (akin to Matlab and scipy's `resample` functions) to impulse.

AngelEzquerra:
@mratsim, I am having trouble to get the gems based correlation work for integers. I'm sure it can be fixed but perhaps it could make sense to merge this improvement in (which makes the performance significantly better) and switch to a gemm based solution on a separate PR?

AngelEzquerra:
> @mratsim, I am having trouble to get the gems based correlation work for integers. I'm sure it can be fixed but perhaps it could make sense to merge this improvement in (which makes the performance significantly better) and switch to a gemm based solution on a separate PR?

Thanks to @Vindaar 's help I was able to make the gemm based convolution algorithm work. I just updated the PR with the changes.

src/arraymancer/tensor/math_functions.nim
```diff
@@ -308,36 +309,135 @@ proc almostEqual*[T: SomeFloat | Complex32 | Complex64](t1, t2: Tensor[T],
...
+    result_len = (f.len + (2*padding) - g.len) div stride + 1
+    kernel_col = g.reshape(1, g.len)
+
+  # Prepare the `result` tensor whose shape must be `[1, N]`,
+  # otherwise the `gemm` call below doesn't do the right thing!
+  result = newTensorUninit[T](1, result_len)
+
+  # # Let's make sure both inputs are contiguous
+  # let f = f.asContiguous()
+  # let g = g.asContiguous()
```
Vindaar, line 407: Can be removed, as it's now further up. :)

Vindaar:
Great work!

PR #652, opened 2024-05-16
Add versions of `toTensor` that take a type as their second argument

AngelEzquerra:
These are convenience functions which are equivalent to calling the regular `toTensor` followed by `asType`. This makes code that uses this relatively common idiom less verbose.


Vindaar:
Yup, agreed these are useful! Have the same in Datamancer for `Column ⇒ Tensor` for the same reason.

PR #653, opened 2024-05-28
Type check accessors `[]`, `[]=` and add `index_select` for multiple values

Vindaar:
This finally makes the `[]`, `[]=` mostly "type safe". Just handing in any type and falling back to the compile time errors of the (mostly internal) procedures like `index_select` is a thing of the past now. If the argument is not allowed, we simply throw a (better) error message telling the user that the type is not valid.

On Nim devel we can also type check `Tensor[T]` (i.e. the inner type), because nowadays doing

```nim
  let validTensorTypes = [getTypeInst(Tensor[int])[1],
                          getTypeInst(Tensor[bool])[1]]
  for t in validTensorTypes:
    if sameType(typ, t): return true
```

works as expected. Nim issue #14021 https://github.com/nim-lang/Nim/issues/14021 should in principle be fixed on devel (but not 2.0.4). Haven't check the Nim issue fully yet though.

This also adds support to not only index a tensor at multiple indices, but also assign multiple at the same time:

```nim
z2[[0, 2]] = [5.2, 1.2]
z2[[0, 2]] = @[5.2, 1.2]
z2[@[0, 2]] = @[5.2, 1.2]
z2[@[0, 2]] = toTensor [5.2, 1.2]
z2[toTensor [0, 2]] = [5.2, 1.2]
z2[toTensor [0, 2]] = @[5.2, 1.2]
z2[toTensor [0, 2]] = toTensor [5.2, 1.2]
```
are all valid now. Previously we could only read with multiple indices.

Note that multi dimensional slicing of arbitrary indices is still not supported.

### DIFF
--- a/src/arraymancer/private/ast_utils.nim
+++ b/src/arraymancer/private/ast_utils.nim
@@ -67,8 +67,7 @@ proc replaceSymsByIdents*(ast: NimNode): NimNode =
     of nnkLiterals:
       return node
     of nnkHiddenStdConv: # see `test_fancy_indexing.nim` why needed
-      expectKind(node[1], nnkSym)
-      return ident($node[1])
+      return inspect(node[1])
     else:
       letsGoDeeper()
   result = inspect(ast)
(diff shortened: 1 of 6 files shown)


AngelEzquerra:
This is great, @Vindaar. Other than the weird git formatting on [3211c4e](https://github.com/mratsim/Arraymancer/pull/653/commits/3211c4e3c12adc26d03fe6f91f8d20d9230fccae)[3211c4e](https://github.com/mratsim/Arraymancer/pull/653/commits/3211c4e3c12adc26d03fe6f91f8d20d9230fccae), that we already discussed offline these changes (and this functionality!) look good to me!

PR #655, opened 2024-06-04
Add "format string" support for tensors

AngelEzquerra:
This comes in 2 forms:
1. Add a version of the pretty function that takes a format string as its input instead of a precision value. Also add a new "showHeader" argument to the original pretty function. - This new version of pretty let's you specify the format string that must be used to display each element. It also adds some new tensor-specific format string "tokens" that are used to control how tensors are displayed (beyond the format of each element).
2. Add a formatValue procedure that takes a tensor as its first input. This makes it possible to control how tensors are displayed in format strings, in the exact same way as if you were using the new pretty(specifier) procedure.

The special, tensor-specific tokens added by this change are:
- "[:]": Display the tensor as if it were a nim "array". This makes it easy to use the representation of a tensor in your own code. No header is shown.
- "[]": Same as "[:]" but displays the tensor in a single line. No header is shown.
- "<>": Combined with the 2 above (i.e. "<>[:]" or "<>[]") adds a header with basic tensor info (type and shape). "<:>" can be used as a shortcut for "<>[:]" while "<>" on its own is equivalent to "<>[]". Can also be combined with "<>||" (see below).
- "||": "Pretty-print" the tensor without a header. This can also be combined with "<>" (i.e. "<>||") to explicitly enable the default mode, which is pretty printing with a header.
- 'j': Formats complex values as A+Bj like in mathematics. Ignored for non-Complex tensors

Note that these are only used to control how the tensors themselves are displayed as a whole and are removed before displaying the individual elements.


Vindaar:
I think this PR could use some other eyes / opinions. Not so much because of the added code (you're welcome to review of course), but more so because of the added syntax. @mratsim @hugogranstrom @Clonkk (feel free to ping anyone else!)

If we decide to add it to arraymancer directly (and not e.g. as a submodule that can be imported `import arraymancer / pretty_printing`), we might want to be more certain that we like the syntax. Any input would be appreciated.

src/arraymancer/tensor/display.nim
```diff
@@ -13,22 +13,160 @@
...
+  ##
+  ## The "format specifier" is similar to those used in format strings, with
+  ## the addition of a few, tensor specific modifiers (shown below).
+  ##
+  ## Inputs:
+  ##   - Input Tensor
+  ##   - specifier: A format specifier similar to those used in format strings,
+  ##                with the addition of a few, tensor specific modifiers which
+  ##                can be combined to achieve different results:
+  ##                - "[:]": Display the tensor as if it were a nim "array".
```
Clonkk, line 63: Does it look at the flattened Tensor or buffer in case of non-contiguous Tensor or Row / Col Major difference ?

src/arraymancer/tensor/display.nim
```diff
@@ -13,22 +13,160 @@
...
+
+proc pretty*[T](t: Tensor[T], specifier: static string = ""): string =
+  ## Pretty-print a Tensor with the option to set a custom format `specifier`
+  ##
+  ## The "format specifier" is similar to those used in format strings, with
+  ## the addition of a few, tensor specific modifiers (shown below).
+  ##
+  ## Inputs:
+  ##   - Input Tensor
+  ##   - specifier: A format specifier similar to those used in format strings,
```
Clonkk, line 60: Shouldn't those be an enum instead of static[string] ?

Clonkk:
Good stuff over all. To me the biggest discussion to be had is about using string as specifier.


For non-mutually exclusive options we could use a set (something done in the std lib).
Maybe, another possibility is to have different proc and pass an enum to do the dispatch in prettyPrint ?

HugoGranstrom:
Thanks for working on this @AngelEzquerra 😄 

I agree with Clonkk that string formats are quite confusing. 
I don't see what the `<>` format would add that a separate `showHeader` parameter couldn't bring. And the different modes (`[]`, `[:]`, `||`) could be replaced with a `printMode` parameter that takes an `enum` as Clonkk suggested. So my suggestion would be a function with this signature:

```nim
type
  PrintMode = enum
    arrayMode
    singleLineMode
    prettyMode

proc pretty*[T](t: Tensor[T], formatString: string, showHeader: bool = true, printMode: PrintModeEnum = prettyMode): string =
    ...
```


Or is it something I'm missing here, some combination of tokens that can't be handled this way?

AngelEzquerra:
Thank you @Vindaar for asking others to comment and thank you for those comments guys! :-)

Just to give more context to other people, the main reason I made this PR is that while the current pretty printing of tensors is nice (and pretty! :-) ), it makes it hard to compare arraymancer's results with numpy's, which uses pretty different format, and, above all, it makes it really hard to use arraymancer's own output in your arraymancer code (e.g. when creating a test or using arraymancer to build a simulator). Another reason is to be able to control the format of the elements themselves (e.g. to use hex format, etc). Finally, I wanted to be able to do all of those while using nim's strformat, which I use extensively and find very useful and nice.

Another comment I'd make is that I tried several options before landing on the current proposal. The first thing I tried is to add an enum argument to `pretty`. The problem with that was that it was not usable with format strings nor let me configure the display of the individual elements. That is why I decided to simply add support for format strings. At first, I tried to use simple letters to indicate that the format should be array-like ("a"), single line array-like ("s") or that it should not have a header ("n"). However, after some thought I got worried that nim's own strformat specifiers (in order to support the control of the format of the individual elements) might evolve and potentially add a use for those letters in the future. That's why I decided to use something that seems much more unlikely to collide with nim's strformat. In particular, I decided to use symbols (and in fact more than one of them). This has the nice side-effect of making the chosen tokens kind of look like the desired output (<> for headers, [] for single-line arrays, [:] more arrays with more than one line and || for the existing table format).

One small tweak that could be done to the current proposal is to force these "tokens" to surround the "element level format". That is, instead of (or perhaps in addition to) adding "[]" to "+06.2f" (i.e. "[]+06.2f"), we could do "[+06.2f]".

src/arraymancer/tensor/display.nim
AngelEzquerra: Using an enum is the first thing I tried but it was not very flexible. Using a string is how I let the user configure the individual element display in the same flexible manner that nim's strformat allows. For example, you can display integers as hex or as binary values, you can force the display of the sign for positive values, you can specify the number of digits in total and on the decimal part or (with nim 2.2) you can ask to use the "math-like" "j" complex display format (which displays complex numbers as `2.3-7.1j`) which is particularly useful for a tensor library in the context of signal processing.

HugoGranstrom:
Yes it is great and something I have also been a bit annoyed at some times that the string representation is a bit odd. 😅 

>  The first thing I tried is to add an enum argument to pretty. The problem with that was that it was not usable with format strings nor let me configure the display of the individual elements.

One doesn't exclude the other. You can specify the array format with an enum and the element format with a string (like in my example). Or was there anything that didn't work out with that approach? :)

AngelEzquerra:
> Thanks for working on this @AngelEzquerra 😄
> 
> I agree with Clonkk that string formats are quite confusing. I don't see what the `<>` format would add that a separate `showHeader` parameter couldn't bring. And the different modes (`[]`, `[:]`, `||`) could be replaced with a `printMode` parameter that takes an `enum` as Clonkk suggested. So my suggestion would be a function with this signature:
> 
> ```nim
> type
>   PrintMode = enum
>     arrayMode
>     singleLineMode
>     prettyMode
> 
> proc pretty*[T](t: Tensor[T], formatString: string, showHeader: bool = true, printMode: PrintModeEnum = prettyMode): string =
>     ...
> ```
> 
> Or is it something I'm missing here, some combination of tokens that can't be handled this way?

That is almost identical to the first thing I tried to do :-)
What that does not let you do (and the reason why I landed on this proposal) is configure the format of the individual elements. It also does not let you use nim's format strings (which I use very often) to print your tensors. With the current proposal you can do something like this:
```nim
echo &"Error the input tensor ({t:[]}) does not have the right rank ({t.rank})"
```
Or something like this:
```nim
echo {t=:+06.8f[]}
```
And get a nice representation of my tensor in which every element has the "+06.f" format _and_ I can copy it into my code because it is valid nim syntax.

Clonkk:
> The problem with that was that it was not usable with format strings nor let me configure the display of the individual elements.


Yeah I understand, enum felt limited too when i thought about it.

In your opinion do you think a set of options would work ? 

´´{HexaFmt, ArrayDisplay, NoHeader}´´ for example would display hexadecimal in array format without the header.

This seems flexible enough while allowing for high cardinality of options and at the same time doesn't require sanitizing input and is resilient to change in the specs (if an option disappears or is split into 2, it will instantly be a compile time error).

Since enums can have a string representation, we could define

´´´
type DisplayOpts = FormatNumbersEnum|HeadersEnum|...
´´´

Since enum can have a string value it shouldn't be too difficult to use the enum in a strformat.

HugoGranstrom:
> What that does not let you do (and the reason why I landed on this proposal) is configure the format of the individual elements.

That is what the `formatString` parameter is for, the format string for the individual elements. I could have been clearer about what I meant :D.  

> It also does not let you use nim's format strings (which I use very often) to print your tensors.

Okay, I think I'm starting to see the problem, it is specifically the `&"{t:[]}"` syntax you are after. I didn't read well enough and thought the updated `pretty` function was the important part here. 😅

AngelEzquerra:
> > The problem with that was that it was not usable with format strings nor let me configure the display of the individual elements.
> 
> Yeah I understand, enum felt limited too when i thought about it.
> 
> In your opinion do you think a set of options would work ?
> 
> ´´{HexaFmt, ArrayDisplay, NoHeader}´´ for example would display hexadecimal in array format without the header.
> 
> This seems flexible enough while allowing for high cardinality of options and at the same time doesn't require sanitizing input and is resilient to change in the specs (if an option disappears or is split into 2, it will instantly be a compile time error).

FYI, strformat strings also give compile time errors because they are all static strings :-) This helped me more than once when I was working on this feature.

> Since enums can have a string representation, we could define
> 
> ´´´ type DisplayOpts = FormatNumbersEnum|HeadersEnum|... ´´´
> 
> Since enum can have a string value it shouldn't be too difficult to use the enum in a strformat.

Sorry, can you explain that a bit more? how would you do that? I'm not sure I understand your proposal.

Clonkk:
> Sorry, can you explain that a bit more? how would you do that? I'm not sure I understand your proposal.

My idea was to use set to pass options to pretty print : 

Typically something like : 

```nim
var A : Tensor[float] = randomTensor(3, 5 ,6, 1.0)
echo a.prettyPrint({optNoHeader, optHexFormat, ... })
```

This make having inclusive options easy and from a doc / api point of view it makes a lot of sense and it is easy to know what's possible and what isn't. It's nice to have but won't solve
the std/strformat problem. 

We can always go for the string specifier and build a way to generate the string specifier from a set of enums (like have ``{optNoHeader, optHexFormat, ... }`` be converted to a specific string specifier at compile time ) so we could have the best of both world.

AngelEzquerra:
> > Sorry, can you explain that a bit more? how would you do that? I'm not sure I understand your proposal.
> 
> My idea was to use set to pass options to pretty print :
> 
> Typically something like :
> 
> ```nim
> var A : Tensor[float] = randomTensor(3, 5 ,6, 1.0)
> echo a.prettyPrint({optNoHeader, optHexFormat, ... })
> ```
> 
> This make having inclusive options easy and from a doc / api point of view it makes a lot of sense and it is easy to know what's possible and what isn't. It's nice to have but won't solve the std/strformat problem.
> 
> We can always go for the string specifier and build a way to generate the string specifier from a set of enums (like have `{optNoHeader, optHexFormat, ... }` be converted to a specific string specifier at compile time ) so we could have the best of both world.

What if we extended the version of `pretty` that doesn't take a format string (the one that takes a precision value) with these new set-based options, but kept the version that takes the format string as is? That way people who don't want to learn about this custom syntax or who don't care about format strings can use the more functional style while those that like format strings can use the new syntax?

Clonkk:
> What if we extended the version of `pretty` that doesn't take a format string (the one that takes a precision value) with these new set-based options, but kept the version that takes the format string as is? That way people who don't want to learn about this custom syntax or who don't care about format strings can use the more functional style while those that like format strings can use the new syntax?

Yes, this is what I had in mind in my last paragraph. But it can be done after this PR once string specifier are merged

Clonkk (approved):

Clonkk:
This seems good enough for me. I think we should merge this whenever CI is green to avoir PR rot

src/arraymancer/tensor/display.nim
```diff
@@ -13,22 +13,160 @@
...
+  ## Pretty-print a Tensor as a "table" with a given precision and optional header
+  ##
+  ## Pretty-print a Tensor with options to set a custom `precision` for float
+  ## values and to show or hide a header describing the tensor type and shape.
+  ##
+  ## Inputs:
+  ##   - Input Tensor.
+  ##   - precision: The number of decimals printed (for float tensors),
+  ##                _including_ the decimal point.
+  ##   - showHeader: If true (the default) show a dscription header
```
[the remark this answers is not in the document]
Vindaar: Fair point, but https://github.com/mratsim/Arraymancer/issues/660 raises a good point about our colMajor support.

Vindaar, line 29:
```suggestion
  ##   - showHeader: If true (the default) show a description header
```

src/arraymancer/tensor/private/p_display.nim
```diff
@@ -14,8 +14,64 @@
...
+  elif specifier.count('|') == 2 or ("[" in specifier and "]" in specifier):
+    return false
+  return true
+
+type tensorDispMode = enum table, multi_line_array, single_line_array
+
+func getTensorDispMode(specifier: string): tensorDispMode =
+  ## Get the display mode based on the special "tensor specifier tokens"
+  let brackets = "[" in specifier and "]" in specifier
+  let colon_brackets = specifier.count(':') == 2
```
Vindaar, line 45: Why do we test for 2 `:` here? Shouldn't it be one?

src/arraymancer/tensor/private/p_display.nim
```diff
@@ -14,8 +14,64 @@
...
+  let flat_brackets = specifier.count('|') == 2
+  if flat_brackets:
+    # Note that "<>||" also implies "table" mode!
+    table
+  elif colon_brackets or
+      (":" in specifier and (brackets or logic_brackets)):
+    # <:> is a shortcut for <>[:]
+    multi_line_array
+  elif brackets or logic_brackets:
+    # <> is a shortcut for <>[] _when_ not combined with ||, ::, [:] or [::]
```
Vindaar, line 56: Huh, `::` is not mentioned in the doc strings, no?

src/arraymancer/tensor/private/p_display.nim
```diff
@@ -152,17 +215,19 @@ func genLeftIdx(axIdx: string, s: string): string =
...
   ## Determines the length of the "largest" element in the tensor after
   ## string conversion. This is to align our output table nicely.
-  when T is SomeFloat:
-    result = t.map_inline((x.formatBiggestFloat(precision = precision)).len).max
-  else:
-    result = t.map_inline(($x).len).max
+  # when T is SomeFloat:
+  #   result = t.map_inline((x.formatBiggestFloat(precision = precision)).len).max
+  # else:
+  #   result = t.map_inline(($x).len).max
```
Vindaar, line 224: This should be removed, no?

src/arraymancer/tensor/private/p_display.nim
```diff
@@ -215,4 +287,136 @@ proc prettyImpl*[T](t: Tensor[T],
...
+  # Remove the separator from the last element
+  result = result[0..^(1+sep.len)] & "]"
+  when T is Complex and "j" in specifier:
+    result = result.replace("(").replace(")")
+
+proc compactTensorDescription[T](t: Tensor[T]): string =
+  ## Describe the tensor in terms of its shape and type (in a "compact" way)
+  ## Only used for array-style printing
+  # Most if not all tensors element types are part of the system or complex
+  # modules so we can remove them from the type without must information loss
```
Vindaar, line 312:
```suggestion
  # modules so we can remove them from the type without much information loss
```

src/arraymancer/tensor/private/p_display.nim
```diff
@@ -215,4 +287,136 @@ proc prettyImpl*[T](t: Tensor[T],
...
+  ## Display a tensor as a multi-line "array"
+  result = t.dispTensorAsSingleLineArray(
+    precision = precision, specifier = specifier, rowSep="\n", showHeader = false)
+  result = indentTensorReprRows(result, t.rank).strip(leading=false)
+  if showHeader:
+    result = t.compactTensorDescription() & ":\n" & result
+
+proc prettyImpl*[T](
+    t: Tensor[T], precision: int, specifier: static string): string =
+  ## Non public implementation of the pretty function
```
Vindaar, line 396:
```suggestion
  ## Public implementation of the pretty function
```
?

Vindaar:
Awesome work and sorry again for the very long delay!

src/arraymancer/tensor/private/p_display.nim
AngelEzquerra: "brackets" is when you use [] while "colon brackets" is when you use :: that's why we must check for a count of 2 colons.

src/arraymancer/tensor/private/p_display.nim
AngelEzquerra: Definitely.

src/arraymancer/tensor/private/p_display.nim
AngelEzquerra: Users are not expected to use the `prettyImpl` function. It is "public" in the sense that it is exported from `p_disp.nim`, but `p_disp.nim` itself is not exported so in the end it should be non-public?

src/arraymancer/tensor/private/p_display.nim
AngelEzquerra: Will fix this typo in a commit taking care of all your comments.

src/arraymancer/tensor/display.nim
AngelEzquerra: Will fix this typo in a commit taking care of all your comments.

PR #657, opened 2024-06-17
Make the output of `ismember` have the same shape as its first input

AngelEzquerra:
Also add documentation for the `ismember` function.


Vindaar:
Nice, thanks!

PR #661, opened 2024-08-08
Add support for running some key `bitops` functions on integer Tensors

AngelEzquerra:
While we already supported most of nim's std/math features in Arraymancer, we did not support any of the std/bitops operators and procedures yet. These are very useful to implement some important algorithms such as gray coding and others.

This commit adds some of the most important std/bitops features. These will soon be used in `impulse` to implement some new algorithms.

I will probably add additional procedures (e.g. setBit, etc.) later, on a separate PR.


Vindaar:
CI failures unrelated. Need to enable `-d:ssl` apparently now for the MNIST download.

Merging.

Issue #660, opened 2024-07-17
Layout (C- or F-) inference leads to incorrect behaviour


darkestpigeon:
Here is a simple example:
```
import arraymancer

let t = arange(0, 2*3)
let w = t.reshape(2, 3).permute(1, 0).flatten()
echo t
echo w

assert t != w
```
The assertion actually fails. The tensors are the same, while you'd expect the `w` tensor to be `[0, 3, 1, 4, 2, 5]`.

The problem lies in the `flatten` (or equivalently `reshape(2*3)`) operation: if `.is_F_contiguous` check passes (and it does in this case), the operation just changes the shape without a copy. In reality we have a non-contiguous C-tensor, so a copy is needed to flatten it. Just because the strides happen to make it look like some contiguous F-tensor, it doesn't become one.

There are two solutions to this problem:
1. Make the layout an attribute of the Tensor object and support both layouts. This way there will be no confusion, but many operations will have to be adjusted.
2. Make all tensors C-tensors (i.e. with rowMajor layout). Actually this is already the case, but checks for F-continuity and layout options need to be eliminated from all operations (`reshape` included), as these are applied erroneously.

Clonkk:
AFAIK, the way to change a tensor layout is through `initTensorMetadata` and not by reshaping which will mostly change strides vector and not actual data.

What I get from this issue :
* Comparaison needs to take strides / shape / col-row major argument 
* We need a dedicated function to swap from row major to col major to avoid those reshape / permute shenanigans

PR #663, opened 2024-08-20
Reshape consistency fix

darkestpigeon:
# What?

Fixing a bug with `reshape`, where it mistakes a non-contiguous C layout for contiguous F layout (see [issue #660](https://github.com/mratsim/Arraymancer/issues/660)).

# Why

A simple illustration why this is a problem:
```
import arraymancer

let x = arange[float](0, 6).reshape(2, 3).permute(1, 0) # has shape [3, 2] and is not C-contiguous (but is F-contiguous)
let y = x.reshape(3*2).reshape(3, 2) # the reshapes shouldn't change the contents

echo x
echo y
doAssert x == y
```
Right now this assertion fails.

# How

The check for whether the tensor is F-contiguous is removed. No-copy reshape is only performed if the tensor is C-contiguous.


mratsim:
@coderabbitai full review

mratsim:
Thank you!

PR #664, opened 2024-08-20
A deleter field (function pointer) for CpuStorageObj

darkestpigeon:
# What

Added a `deleter` field to `CpuStorageObj`, which will be called on `memalloc` field instead of `deallocShared` if not `nil`.

# Why

This change allows the implementation of memory sharing with other objects. For example, we can acquire memory from some opaque type
```

let x: SomeOpaquePointer = funcFromC(...)
var t = newTensorUninit[float]()

t.storage.cpuStorageFromBuffer(
  getData(x), cast[pointer](x),
  proc(x: pointer) {.noconv.} = let x = cast[SomeOpaquePointer](x); destroyFromC(x))
```

or we can wrap the tensor's memory into a shared pointer, and make `deleter` to delete a single copy of this pointer. This will facilitate sharing memory owned by tensors between threads.

# How

A field `deleter: proc(memalloc: pointer) {.noconv, gcsafe, raises: [].}` was added to the `CpuStorageObj`. This field is then used in the destructor.

### DIFF
--- a/src/arraymancer/laser/tensor/datatypes.nim
+++ b/src/arraymancer/laser/tensor/datatypes.nim
@@ -40,9 +40,10 @@ type
   CpuStorageObj[T] {.shallow.} = object
     # Workaround supportsCopyMem in type section - https://github.com/nim-lang/Nim/issues/13193
     when T is KnownSupportsCopyMem:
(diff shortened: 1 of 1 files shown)


mratsim:
@coderabbitai full review

mratsim:
Thank you!

PR #665, opened 2024-09-01
Update arraymancer dependencies

AngelEzquerra:
The dependency with nim was very out of date (which caused problems with nimble and nim's language server). I'm also taking the chance to make small bumps to some of the other dependencies.


Vindaar:
Thanks!

PR #668, opened 2024-10-19
Fix cuda

lilkeet:
this pr attempts to fix the cuda compilation. if u can configure urself into a position that a cuda compiler likes, all but one of the cuda unit tests compiles and runs successfully.


mratsim (approved):
Thank you!

PR #671, opened 2024-11-09
Make the DSL interpret sym nodes as ident nodes

fabricio-p:


fabricio-p:
I suspect that the usually `nnkIdent` nodes became sym nodes because I tried invoking the `network` macro inside a template. There could be a greater issue at play here (in the compiler) but I digress

Vindaar:
> I suspect that the usually `nnkIdent` nodes became sym nodes because I tried invoking the `network` macro inside a template. There could be a greater issue at play here (in the compiler) but I digress

No, that's precisely what would happen in that context. It will only be `ident` if it is really in a fully untyped context.

Vindaar:
Thanks, looks good to me!

PR #674, opened 2025-02-15
make typedesc arg of `init` procs not require generic param

metagn:
A fix to https://github.com/nim-lang/Nim/issues/24688 could imply that typedesc args with types like `Foo[T]` should not match a type `Foo`, especially if the `T` is inferred. `init` procs are usually called as such i.e. `init(ctx, NetName)`, so their parameters are given a type `typedesc[NetName]` rather than `typedesc[NetName[T]]`. Hopefully this is not too much of an incompatibility.

### DIFF
--- a/src/arraymancer/nn/layers/conv2D.nim
+++ b/src/arraymancer/nn/layers/conv2D.nim
@@ -138,7 +138,7 @@ type
 
 proc init*[T](
   ctx: Context[Tensor[T]],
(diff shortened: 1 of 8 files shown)


metagn:
Tested in Nim CI in https://github.com/nim-lang/Nim/pull/24690, seems to work, CI here seems to be broken for other reasons

Issue #677, opened 2025-02-23
Cuda compilation fails on Nim 2.2.2


lilkeet:
Compilation when using cuda fails on the most recent Nim stable release with: `/src/arraymancer/tensor/backend/cuda.nim(36, 6) Error: cannot bind another '=destroy' to: CudaTensorRefTrackerObj; previous declaration was constructed implicitly`

I think caused by [this pr](https://github.com/nim-lang/Nim/pull/24425)?? If so, i guess it was never supposed to work lol.

Should be an easy fix, maybe.

mratsim:
Closed by #677

PR #678, opened 2025-02-23
Fix cuda on 2.2

lilkeet:
Fixes issue #677 by updating cuda memory management to standard inserted `=destroy` calls.


mratsim:
Thank you!

PR #684, opened 2025-11-10
Fix #682

### LINKED ISSUE #682: Creating an identity matrix crashing program
I just tried to create an identity matrix and the program keeps crashing:

```
import arraymancer
var b = eye[float](3, 3)
echo b
```

or 

```
import arraymancer
var b = identity[float](3)
echo b

```
The output:

```
/home/~/tester.nim(12, 24) template/generic instantiation of `identity` from here
/home/~/Arraymancer/src/arraymancer/linear_algebra/special_matrices.nim(238, 16) template/generic instantiation of `diag` from here
/home/~/Arraymancer/src/arraymancer/linear_algebra/special_matrices.nim(226, 9) template/generic instantiation of `set_diagonal` from here
/home/~/Arraymancer/src/arraymancer/linear_algebra/special_matrices.nim(172, 26) Error: type mismatch
Expression: &"Diagonal index ({k=}) exceeds input matrix height ({a.shape[0]})"
  [1] "Diagonal index ({k=}) exceeds input matrix height ({a.shape[0]})": string

Expected one of (first mismatch at [position]):
[1] func `&`(a, b: DynamicStackArray): DynamicStackArray
[1] func `&`[T](a: DynamicStackArray[T]; value: T): DynamicStackArray[T]
[1] proc `&`(x, y: char): string
[1] proc `&`(x: char; y: string): string
[1] proc `&`[T](x, y: sink seq[T]): seq[T]
[1] proc `&`[T](x: sink seq[T]; y: sink T): seq[T]
[2] proc `&`(x, y: string): string
  missing parameter: y
[2] proc `&`(x: string; y: char): string
  missing parameter: y
[2] proc `&`[T](x: sink T; y: sink seq[T]): seq[T]
  missing parameter: y
```

lilkeet:
Solves issue #682.
The `&` proc was used before the `std/strformat` module was imported.
The test suite missed this because it imported the proc itself.

### DIFF
--- a/src/arraymancer/linear_algebra/special_matrices.nim
+++ b/src/arraymancer/linear_algebra/special_matrices.nim
@@ -129,6 +129,7 @@ proc diagonal*[T](a: Tensor[T], k = 0, anti = false): Tensor[T] {.noInit.} =
   ##      - anti: If true, get the k-th "anti-diagonal" instead of the k-th regular diagonal.
   ## Result:
   ##      - A copy of the diagonal elements as a rank-1 tensor
+  bind `&`
   assert a.rank == 2, "diagonal() only works on matrices"
   assert k < a.shape[0], &"Diagonal index ({k=}) exceeds the output matrix height ({a.shape[0]})"
   assert k < a.shape[1], &"Diagonal index ({k=}) exceeds the output matrix width ({a.shape[1]})"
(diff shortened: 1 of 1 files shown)


mratsim:
The issue is that this will happen for every single generic proc, and all procs are genetic on `Tensor[T]` in the library. It's a Nim limitation that is documented in `strformat` though it only mentions template expansions, generics are the same: https://nim-lang.org/docs/strformat.html#limitations

